PR #6, opened 2018-10-16
JWT authenticated endpoints

JPrevost:
#### What does this PR do?

- provides a `/ping` endpoint that works with or without authentication.
- provides and `/auth` endpoint that takes an existing / confirmed users email and password in headers and returns a JWT.
- provides a `/search` endpoint that only accepts valid JWT and returns dummy data.

This code does not provide any mechanism to server side revoke tokens. We should discuss the need for that and implement whatever we determine is appropriate.

Blog much of this came from:
https://medium.com/@goncalvesjoao/rails-devise-jwt-and-the-forgotten-warden-67cfcf8a0b73

Starting point for background info on revocation:
http://waiting-for-dev.github.io/blog/2017/01/23/stand_up_for_jwt_revocation/

Gem I chose not to use because it was poorly documented in terms of how to actually use it:
https://github.com/waiting-for-dev/devise-jwt

I feel comfortable not using it as the solution here is small and understandable, but I wanted to share some of the process I went through in reaching this solution.

If you register and login to the UI, you'll see a curl statement to test the endpoint after your confirm your registration and login. Test it with the command provide and with a wrong token.

You can also request a token via API (after registering in the ui) via:

```
curl -H "Content-Type: application/json" -H "X-User-Email: YOUREMAIL" -H "X-User-Password:YOURPASSWORD" -X GET http://localhost:3000/auth
```

- https://mitlibraries.atlassian.net/browse/DIP-158

NO

YES


app/controllers/auth_controller.rb
```diff
@@ -0,0 +1,22 @@
+class AuthController < ApplicationController
+  respond_to :json
+  before_action :ensure_json!
+
+  def auth
+    request.format = :json
+    email = request.headers['X-User-email']
```
gravesm: Is there a reason you didn't just use basic auth here?

JPrevost: No. I could rework that if you feel it's better to use that standard.

JPrevost: I'll change this to using basic auth.

JPrevost:
@gravesm 
https://github.com/MITLibraries/timdex/pull/6/commits/a2cb3f9bf2a2deeb495c518c75f60e4ba78c1bb2 takes your suggestion to use basic auth. Look at all that code disappear ;) 🍻

PR #13, opened 2018-10-30
AWSv4 Signing Option for ElasticSearch

JPrevost:
#### What does this PR do?

Adds AWSv4 Signing Option for ElasticSearch

Setting the ENV as described in the readme, you should be able to connect to either our AWS ES instance or your localhost (or probably other endpoints but I didn't test that).

- https://mitlibraries.atlassian.net/browse/DIP-164

NO

YES

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -4,13 +4,11 @@ class SearchController < ApplicationController
   before_action :authenticate_user!, except: 'ping'
 
   def search
-    client = Elasticsearch::Client.new log: false
-    @results = client.search(q: params[:q])
+    @results = Timdex::EsClient.search(q: params[:q])
   end
 
   def record
-    client = Elasticsearch::Client.new log: false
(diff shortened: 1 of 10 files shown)


app/controllers/search_controller.rb
```diff
@@ -1,15 +1,16 @@
 class SearchController < ApplicationController
+  require 'faraday_middleware/aws_sigv4'
   respond_to :json
   before_action :ensure_json!
   before_action :authenticate_user!, except: 'ping'
 
   def search
-    client = Elasticsearch::Client.new log: false
+    client = configure_elasticsearch
```
gravesm: It looks to me like you are creating a new ES client for every request. I would think this is probably not what we want. I'm not sure how Rails typically handles this, but I would guess client creation might go in an initializer or something?

JPrevost: Yup. You are correct. I'll adjust (stupid following docs instead of thinking about what is going on).

PR #19, opened 2018-11-15
namespaced api and cors

JPrevost:
#### What does this PR do?

Moves to version namespaced API to allow for easier future breaking changes with less effort.

Also adds CORS (sorry for two in one PR but I ran into issues in some documentation work that needs CORS so I want to slide it in now).

https://paweljw.github.io/2017/07/rails-5.1-api-app-part-3-api-versioning/

NO

YES


app/views/static/index.html.erb
```diff
@@ -3,5 +3,5 @@
 <p>This is a splash page. It is amazing.</p>
 
 <% if user_signed_in? %>
-  curl -X GET --header 'Authorization: Bearer <%= JWTWrapper.encode({ user_id: current_user.id }) %>' '<%= search_url(q: 'whatever')%>'
+  curl -X GET --header 'Authorization: Bearer <%= JWTWrapper.encode({ user_id: current_user.id }) %>' '<%= api_v1_search_url(q: 'whatever')%>'
```
hakbailey: Is there any reason to/not to variableize (not a word) the API version here rather than hard coding it in?

JPrevost: This is totally a sample code thing we'll pull out once we make a real splash page and docs.

PR #23, opened 2018-11-19
Allows querying of specific es index

JPrevost:
#### What does this PR do?

Previously, the search queried all indexes but the retrieve was on a fixed index. This presented problems when multiple indexes existed as search results would lead to non-retrieable full records.

NO

NO


JPrevost:
@gravesm @hakbailey I'm merging as I have work that gets more complicated to PR if this doesn't merge. I'm happy to discuss any concerns once you return from vacation(s).

PR #39, opened 2019-01-08
Rails changed behavior in a dot release and broke this

JPrevost:
#### What does this PR do?

Updates Lograge timestamp to avoid change in recent Rails dot release.

- https://mitlibraries.atlassian.net/browse/DIP-222

NO

NO


JPrevost:
@gravesm You'll probably want to update this as well eventually before updating to the latest Rails:
https://github.com/MITLibraries/geoweb/blob/master/config/initializers/lograge.rb#L10

PR #40, opened 2019-01-08
Redirect to preferred domain if it is set in ENV

JPrevost:
#### What does this PR do?

Setting `PREFERRED_DOMAIN` will redirect any requests to any other domain that hit this app to redirect to the root of the PREFERRED_DOMAIN.

This is useful to prevent inadvertent access to herokuapp.com domains in production, as well as handling any legacy domains if they exist (which they don't in this app...at least yet).

If you set the `PREFERRED_DOMAIN` ENV and then try to access the PR build, you'll get redirected to whatever you set... which will only be partially testing this as you won't end up on the PR build. So yeah, it mostly needs to be fully confirmed in production.

- https://mitlibraries.atlassian.net/browse/DIP-202

NO

YES

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,6 +38,7 @@ end
 group :test do
   gem 'capybara', '>= 2.15'
   gem 'chromedriver-helper'
+  gem 'climate_control'
   gem 'coveralls', require: false
   gem 'selenium-webdriver'
   gem 'timecop'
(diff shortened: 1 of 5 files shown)


JPrevost:
@gravesm We totally use this in other projects and this was copy / pasted over here. Good idea considering using a gem. I'll add that to a maintenance days task to look into (eventually).

PR #55, opened 2019-02-06
Resolve PR Build Errors

JPrevost:
#### What does this PR do?

We've had PR build errors forever on this app. The apps work fine but they always throw an error in the logs. I believe this is why... we don't use workers on this app but told it to spin one up. Ooops.


JPrevost:
FYI: This indeed built in Heroku with no errors.

PR #102, opened 2019-05-08
Anon

JPrevost:
#### What does this PR do?

Allows anonymous access to previously JWT restricted endpoints at a configurable rate (for demo purposes 10 every 5 minutes).

JWT authenticated requests bypass this restriction and work as they always did.

Incomplete
- [x] update public documentation to explain the new ability to fully use the API without registration for most use cases.

Our goal with TIMDEX! is to allow anyone to identify MIT content they are interested in. Locking this behind an access token seems counterproductive to that goal. However, we to keep the tokens in place as a way to requests unmetered access "just in case" we need it. Our intent is to configure this with enough requests in anonymous mode that most people should not need to use tokens.

Without authenticating, request a search endpoint more than 10 times in 5 minutes. You should eventually trip the counter and get a 429 response back. Waiting a few minutes should allow 10 more requests... or, authenticating should allow immediate access and not trip the throttle at all (although keep in mind once you trip it you can't auth until it resets because auth is an endpoint that is throttled).

```
curl -i -X GET "https://timdex-stage-pr-102.herokuapp.com/api/v1/search?q=popcorn" -H "accept: application/json"
```

- https://mitlibraries.atlassian.net/browse/DIP-260

NO

YES


config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,114 @@
...
+  # If any single client IP is making tons of requests, then they're
+  # probably malicious or a poorly-configured scraper. Either way, they
+  # don't deserve to hog all of the app server's CPU. Cut them off!
+
+  # Throttle all requests by IP
+  #
+  # Key: "rack::attack:#{Time.now.to_i/:period}:req/ip:#{req.ip}"
+  throttle('req/ip',
+           limit: (ENV['REQUESTS_PER_FIVE_MINUTES'] || 500).to_i,
+           period: 5.minutes) do |req|
```
gravesm: Five minutes is kind of harsh. Also, not critical, but I feel like maybe the period could be an env var, too? That way we can adjust without having to do a redeploy.

JPrevost: Hmm. I'm not sure I understand. The 5 minutes gives the ability to have a burst of activity during a 5 minute window. For example, 100 requests per minute would get throttled at 101 requests in one minute. But 500 requests in 5 minutes, the same total requests allowed over 5 minutes, would not throttle that 101st request. It would allow until the 501st request over 5 minutes.

JPrevost: However, I do agree that period should be an ENV to allow for this to be adjustable in production without code changes. 👍

gravesm: I guess my reasoning is that the point of throttling anonymous access is to make sure we still retain some control over server resources. A five minute cooldown seems punitive.

JPrevost: Interesting distinction. I see your point, but I was thinking it was less punitive to apply the throttle less frequently by allowing for bursts. Either way, I think our intent is to try to set the level high enough to rarely throttle anyone that isn't being either kinda dumb or intentionally abusive. I'll be making the period ENV configurable so we can setup reports in our logging to look for this happening in production and adjust until we feel we've got the right numbers for both values. 👍

gravesm: Yeah, good idea to check the logs to see how often throttling happens.

PR #191, opened 2019-10-25
Docker compose to orchestrate es and load sample data

JPrevost:
#### What does this PR do?

The included docker compose allows for easy setup of elasticsearch with a way to load sample data (and run arbitrary Mario commands).

Not yet includes is actually running TIMDEX! in Docker but that will be included soon.

Follow the instructions in the readme and confirm you end up with persistent ElasticSearch and can load the sample data.

Bonus points for also being able to run an arbitrary Mario command such as "indexes"

NO

NO


docker-sampledata.yml
```diff
@@ -0,0 +1,20 @@
+version: '3.7'
+services:
+  aleph:
+    image: mitlibraries/mario:aspace
```
JPrevost: This and the one below are pinned to the `aspace` tag of Mario until that branch lands and then we should pin back to master.

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ruby:2.6
```
gravesm: No reason you can't use the Debian base image, but I'm running the alpine image for GeoWeb in production and it seems to do fine.

docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+      - cluster.name=docker-cluster
+      - bootstrap.memory_lock=true
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+    volumes:
+      - type: volume
+        source: timdex_esdata
```
gravesm: Compose will already prefix the volume name with the current working directory, so this volume ends up being named `timdex_timdex_esdata`. It still works fine as is, it's just a little redundant. I'd just name it `esdata` and then also change the name in the top level `volumes`. Or you can leave it.

PR #384, opened 2020-11-19
Add collections field to data model

hakbailey:
#### What does this PR do?

Updates the search controller and json builders to include a collections field. Adds both a filter and an aggregation for
the new field, and adds it to the base (brief) record view. Updates the API spec to reflect new field, parameter, and
aggregation.

Fully implementing this depends on this field existing in the Mario code (a corollary PR is being submitted simultaneously to that repo), ingesting the DSpace metadata to ES on staging to ensure the collections field is indexed, and pointing the merged TIMDEX code to Mario staging to ensure it all works together.

Adding this field is important for the DSpace ingest parsing recently added to Mario, as Collection is a new metadata field we hadn't encountered in previous sources.

`make up` to get local instances of TIMDEX and Mario running, then from a Mario folder _using the [add_collections_field branch
](https://github.com/MITLibraries/mario/tree/add-collections-field)_ run `mario ingest -t dspace --auto fixtures/dspace_samples.xml`. View query results in the browser, collections field and aggregation should appear. Filtering with `http://localhost:3000/api/v1/search?q=mit&collection[]=mit%20libraries` should also work.

https://mitlibraries.atlassian.net/browse/DISCO-35

NO

NO


app/models/search.rb
```diff
@@ -128,6 +129,11 @@ def filter_single(param, field)
   # https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html
   def aggregations
     {
+      collections: {
```
JPrevost, line 132: FYI: I checked this code against timex-prod and confirmed it doesn't break if we release this before we Elasticsearch has the updated model. It just returns an empty array here like:
``` 
"collection": [],
```

JPrevost (commented):
If you can bump the version in the spec we should be good.

Nothing in this changeset breaks with the current production ES so we once that small change is made we should be good to tag a release and promote to production and once Mario is released and indexes update they'll start having this data available.

JPrevost (approved):

PR #411, opened 2021-01-08
Updates returned order of aggregations

JPrevost:

* The order was rather arbitrary to begin with, and we decided to make
  the order match at least one real world use case, which is our own
  DISCO UI
* Updates nokogiri (unrelated to this work but it's a security patch so it should be updated)

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/DISCO-221

* This returns the aggregations in the order our DISCO UI currently
  desires. Future changes to the display order in any UI should probably
  be handled in the UI itself, but since the initial order was not for a
  specific purpose this was both easier and fine.

NO

YES


jazairi (approved):
Confirmed that TIMDEX returns aggregations in the desired order. :shipit:

PR #500, opened 2022-02-17
Update openapi.json

MaceWindu:
fix broken json document

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -442,13 +442,13 @@
                 "items": {
                   "type": "string"
                 }
+              },
               "collections": {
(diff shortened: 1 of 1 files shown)


JPrevost:
The existing JSON is valid. Can you confirm what issues you noticed that led to you open this Pull Request?

JPrevost:
Oh wait, I misspoke and validated _your_ updated JSON which was valid. The repo is indeed broken. I'll take a look to ensure the JSON you provided is both valid and correct for the spec. Thanks!

JPrevost:
Thanks for the fix! Appreciate it.

PR #504, opened 2022-02-23
Provides end user documentation via GitHub pages

JPrevost:
Generating on-the-fly using Swagger UI and the version of the api spec
in the main branch of the repository.

We will want to consider options to provide more control of which
version of the API spec we use in case we merge changes into main prior
to launching the changes into production but we currently need to get
this solution into production as our current solution was deleted by
the vendor with no notice.

- https://mitlibraries.atlassian.net/browse/ENGX-140

NO

Not directly, but our docs now have an upstream CDN dependency.


JPrevost:
FYI: the docs build and push to `gh-pages` that shows up in Actions for this is not part of this branch or repo and is part of a separate idea I was pursuing. I suspect it'll go away for future builds, but if not I'll figure out how to fix the Actions (the action that ran is not part of Main or this branch so it's definitely a bug).

JPrevost:
You can see the solution in action at https://mitlibraries.github.io/timdex/ (it currently deploys the Pages from this branch but once this merges I'll change to build Pages from main).

matt-bernhardt (approved):
This all seems fine to me, and I philosophically like hosting our self-generating documentation within GitHub Pages over having it on a third-party vendor (given our recent experience in particular). The various example calls in different languages were a nice touch, but not worth everything else.

I've not thoroughly inspected the generated documentation, but what is there seems generally usable and thorough given what's in `openapi.json`

PR #506, opened 2022-03-23
update to ruby 2.7.5

JPrevost:
#### What does this PR do?

updates to ruby 2.7.5

NO

YES


matt-bernhardt (commented):
I'm able to get the front page and the graphql playground running locally after installing 2.7.5, but I think I'm missing the right env / local elasticsearch to confirm everything fully works.

The tests pass, and the ping route works locally.
On the review app, I see the playground and the search route returning records.

Based on this, I think this is okay to merge - but if you'd like me to confirm local functionality I'll need to work with you to get the missing pieces in place.

JPrevost:
@matt-bernhardt as we'll be doing more dev work on this app soon we'll get you setup for local dev on it soonish. For now, I believe you confirmed the app worked as expected in the PR build and CI which were both running the updated Ruby so I think you can approve this PR based on that.

PR #511, opened 2022-04-20
Initial opensearch backed v2 endpoint

JPrevost:

* Being able to connect to OpenSearch from TIMDEX API is a building
  block towards implementing the v2 API

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/RDI-27

* creates a toggleable v2 version of the v1 endpoints
* passes clients into the endpoints to allow v1 and v2 to use different
  clients while (for now) sharing most of the querying logic

* DockerFile, composer, and makefile are not yet updated to reflect
  being backed by both ES and OS at the same time.
* GraphQL is only available on the V1 endpoint
* The v2 endpoint this introduces is largely the same spec as v1 and
  does not reflect the future needs of the v2 endpoint. It essentially
  allows v1-like functionality to be backed by opensearch. We have
  additional tickets open to plan and implement the actual v2 endpoint.

NO

YES


jazairi (changes requested):
I think this approach makes sense, and it should make it easier for us to deprecate v1 should we want to do that down the road. I don't think that some of the CodeClimate suggestions are applicable here (e.g., the DRY warnings), but I would recommend cleaning up the minor stuff (trailing whitespace, empty lines, `instance_of?`, parenthesis) before we merge this.

PR #512, opened 2022-04-22
Update v2 record to map core datacite fields

JPrevost:

* We changed some field names and added some additional fields in the v2
  data model. This brings some core fields up to date.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/RDI-93

* Updates json records to v2 spec (for a small number of fields)

* removes all non dataverse core fields from v2. We'll add them back
  in but it was difficult to keep track of what had already been
  updated when we had some legacy and some new mappings.

NO

NO


jazairi (approved):
:shipit: pending rubocop fixes.

PR #513, opened 2022-05-02
Return remaining V2 fields

jazairi:

The data model includes fields that are not yet returned in V2
of the API.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/RDI-89

This returns the additional V2 fields.

* These fields have been added to the base set of fields, but this
may require modification once we determine the extended set.
* The following fields included in this commit are not required for
RDI: `contents`, `call_numbers`, `holdings`, `literary_form`,
`numbering`, `physical_description`, `publication_frequency.
* Cassettes have been regenerated.
* The `id` field has been changed to `timdex_record_id`, as they
seem to be the same (`timdex_record_id` is listed as a new field
in V2 of the data model). I'd like the reviewer to confirm whether
this is an acceptable approach, or if `id` should remain as mapped
from `timdex_record_id` wih `timdex_record_id` returned as a separate
field or not returned at all.

Some of the new fields will be physical by calling the search endpoint o the V2 API (however, not all of them are represented in our test data yet).

NO

NO


jazairi:
This branch has been deployed to TIMDEX staging.

matt-bernhardt:
I've gone through the commit, confirmed that the tests still pass, and compared the current state of `_base.json.builder` to the v2 spec spreadsheet. I can confirm that every field in the spec is now listed in the base partial (and the base partial introduces no field not listed in the spec).

I'm not sure about the `timdex_record_id` question, though. I understand that breaking changes like renaming this field are permissible, but I'm not sure whether user expectations for an `id` field are strong enough to justify including this value twice (first as `id` and then again as `timdex_record_id`). It is hard for me to anticipate another possible returned ID that could result in another id - like a `foo_id` field and also a `timdex_record_id`, both of which would then be duplicated as just `id`.

I'm going to try and sleep on this, and come at the question tomorrow morning with a fresh mind. I'm posting this comment now to remind future-me of where I left off.

Questions in my mind right now include:
- Should _base.json.builder` alphabetize earlier fields? I see the new fields have been added in alpha order, but the earlier batch were not.
- I'm not sure I follow the rationale behind the changes in the cassettes, but this is more likely my end-of-day brain being generally confused.

test/vcr_cassettes/v2_expired_token.yml
```diff
@@ -84,50 +84,5 @@ http_interactions:
         Jay)"]}],"timdex_record_id":"mit:archivesspace:VC.0002","title":"Charles J.
         Connick Stained Glass Foundation Collection"}}]},"aggregations":{"languages":{"doc_count_error_upper_bound":0,"sum_other_doc_count":0,"buckets":[{"key":"english","doc_count":1}]},"content_type":{"doc_count_error_upper_bound":0,"sum_other_doc_count":0,"buckets":[]},"collections":{"doc_count_error_upper_bound":0,"sum_other_doc_count":0,"buckets":[]},"subjects":{"doc_count_error_upper_bound":0,"sum_other_doc_count":0,"buckets":[]},"content_format":{"doc_count_error_upper_bound":0,"sum_other_doc_count":0,"buckets":[]},"literary_form":{"doc_count_error_upper_bound":0,"sum_other_doc_count":0,"buckets":[]},"source":{"doc_count_error_upper_bound":0,"sum_other_doc_count":0,"buckets":[]},"contributors":{"doc_count":1,"contributor_names":{"doc_count_error_upper_bound":0,"sum_other_doc_count":0,"buckets":[{"key":"connick,
         charles j. (charles jay)","doc_count":1}]}}}}'
-  recorded_at: Thu, 07 Apr 2022 14:00:00 GMT
```
JPrevost, line 87: Yeah so... this is currently dumb but all of the cassettes need these two calls in them. If you run the tests locally enough times, you'll see sporadic failures on each one. What seems to be happening is during initialization of the OpenSearch client, it makes a check for compatibility. This happens on the first test that uses the client and then the rest don't do it. For stability reasons, we need to allow that call in each cassette. Alternatively, we could figure out how to disable that Client check.

I approached this when I was making cassettes by running each test individually until it generated with both HTTP calls in the cassette and then moving on to the next one. It's not really sustainable to do that so at some point we need a better solution but for now that should work I think since we still have a smallish number of cassettes affected by this.

app/views/api/v2/search/_base.json.jbuilder
```diff
@@ -1,9 +1,30 @@
-json.id result['timdex_record_id']
+json.timdex_record_id result['timdex_record_id']
```
JPrevost, line 1: @hakbailey I seem to recall having a conversation around wanting to retain `id` as the user facing version of of `timdex_record_id`. Did we agree that `id` was fine for `timex-api` even if the data model wanted to use `timdex_record_id` or was there a strong desire to change what we refer to the id as outwardly?

hakbailey: This is a case where I wish we could get some user feedback because I'm not sure what the best answer is. `timdex_record_id` is certainly more clear and accurate, especially given that most if not all records will have other IDs (DOIs, ISBNs, OCLCs, etc.) in the `identifiers` field, and those will have a `kind` value to indicate what kind of ID they are. Having an additional `id` field with no indication of what kind of ID it is seems potentially confusing, although we can and should document it in the spec. But I can also see an argument that `id` is a generally well-understood term when dealing with records of all sorts, and users might expect that. Do other folks have strong feelings one way or another?

JPrevost: I think it's pretty common practice to have `id` signify the internal id for whatever you are working with. If we were going to allow multiple IDs in a single route, we'd want to clarify.

For instance `/api/v1/record/:id` is super clear and `/api/v1/record/:timdex_record_id` doesn't feel more clear as all it does is repeat information that is already implied (I'm using timdex and want a specific record by id).

JPrevost: I suggest for now we move back to `id` for this and include evaluating this as part of our initial pathfinder work when we reach out to DLS/Ben. Until we publish v2 fully it will remain easy to swap this out if we find this is confusing.

JPrevost: I suggest for now we move back to `id` for this and include evaluating this as part of our initial pathfinder work when we reach out to DLS/Ben. Until we publish v2 fully it will remain easy to swap this out if we find this is confusing.

JPrevost (approved):

PR #514, opened 2022-05-16
V2 graphql

JPrevost:
This forks the Search model to OpenSearch and connects GraphQL to use that new OpenSearch model.

Please see the commit message for the most recent commit for much more context and side effects of this work.

- Using the built in GrapiQL playground, note the deprecated fields and how they can still be used but throw warnings.
- Retrieve new fields and deprecated fields

- https://mitlibraries.atlassian.net/browse/RDI-101

NO

YES


jazairi (approved):
I won't pretend to grok every detail of this, but based on my understanding of the change set, this seems like a valid solution. I think that once we load the legacy data, we should plan to remove the FlipFlop feature and V1 stuff (assuming I'm correctly interpreting that that will be possible). Aside from making the test suite easier to run, it should make this more maintainable in the long run. Based on my experience doing something similar with Bento, it shouldn't be too painful.

Thanks for taking this on and sharing the context around it. :shipit:

jazairi:
I should add, the test coverage drop doesn't concern me, since most if not all of it seems to be related to the deprecations and V1 side of conditionals, both of which will hopefully be dropped when we remove this in the future. Likewise, many of the flags CodeClimate raised will be lowered as a result of that same maintenance.

PR #516, opened 2022-05-19
Configure Opensearch model to search single fields

jazairi:

The search model needs to be able to target any of the fields defined
as single-field searchable in v2.0 of the data model.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/RDI-102

This enables matching the following fields in the Opensearch model:

* citation
* contributors
* funding_information
* identifiers
* locations
* subjects

(Note that title was already made searchable as part of a previous
commit.)

* These fields are not yet enabled in GraphQL, so you'll need to use
REST to confirm this behavior.
* While this allows us to search a single nested subfield, it does not
it does not allow for searching multiple nested subfields. For example,
searching contributors should search both 'value' and 'identifier', but
right now we can only search 'value'.

NO

NO

### DIFF
--- a/app/models/opensearch.rb
+++ b/app/models/opensearch.rb
@@ -31,6 +31,7 @@ class Opensearch
 
   def multisearch
     return unless @params[:q].present?
+
     [
       {
         prefix: {
@@ -73,14 +74,14 @@ class Opensearch
         }
       }
     end
+    match_single_field(:citation, m)
+    match_single_field(:title, m)
 
-    if @params[:title].present?
-      m << {
-        match: {
-          title: @params[:title].downcase
-        }
-      }
-    end
+    match_single_field_nested(:contributors, m)
+    match_single_field_nested(:funding_information, m)
+    match_single_field_nested(:identifiers, m)
+    match_single_field_nested(:locations, m)
+    match_single_field_nested(:subjects, m)
     m
   end
 
@@ -197,5 +198,43 @@ class Opensearch
         }
       }
     }
+  end
+
+  private
+
+  def match_single_field(field, match_array)
+    return unless @params[field]
+
+    match_array << {
+      match: {
+        field => @params[field].downcase
+      }
+    }
+  end
+
+  def match_single_field_nested(field, match_array)
+    return unless @params[field]
+
+    match_array << {
+      nested: {
+        path: field.to_s,
+        query: {
+          bool: {
+            must: [
+              { match: { "#{field}.#{nested_field(field)}": @params[field].downcase } }
+            ]
+          }
+        }
+      }
+    }
+  end
+
+  # For most nested fields, we only care about 'value'; this handles the exceptions to that rule.
+  def nested_field(field)
+    if field == :funding_information
+      'funder_name'
+    else
+      'value'
+    end
   end
 end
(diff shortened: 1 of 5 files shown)


app/models/opensearch.rb
```diff
@@ -66,21 +67,14 @@ def multisearch
 
   def matches
     m = []
-    if @params[:q].present?
-      m << {
-      multi_match: {
-        query: @params[:q].downcase
-        }
-      }
-    end
```
JPrevost, line 75: I think right now `q` is only matching in titles and contributors from the `multimatch` block. This entire `if` block, which confusingly is a `multi_match` but done in the `match` block (I promise it makes sense!) needs to be added back in.

You can see this by connecting this branch to the dev1 open search and entering this GraphQL Query:
```
{
  search(searchterm: "Standards") {
    hits
    records {
      title
      subjects {
        kind
        value
      }
    }
  }
}
```

Note only one hit. Add the lines back in, rerun the search, and you'll see 6 hits and that there were matches this time in additional fields that should be searched when doing keyword searches. The method naming is definitely not helpful here and we can consider how to better address that as we work more with this together.

app/models/opensearch.rb
```diff
@@ -66,21 +67,14 @@ def multisearch
...
+    match_single_field(:citation, m)
+    match_single_field(:title, m)
+
+    match_single_field_nested(:contributors, m)
+    match_single_field_nested(:funding_information, m)
+    match_single_field_nested(:identifiers, m)
+    match_single_field_nested(:locations, m)
+    match_single_field_nested(:subjects, m)
```
JPrevost, line 77: I like this abstraction.

JPrevost (changes requested):
Solid approach to the problem. I noted a block of code that needs to be added back in and then I think this works for what we need now and if we understand more later about how we want to target nested fields later we can adjust.

app/models/opensearch.rb
jazairi: @JPrevost Ha, I knew that was there for a reason. I added it back in and confirmed the correct behavior by successfully searching metadata in non-title/contributors fields with `q`.

I didn't make a regression test for this because it seemed like it would require generating a ton of cassettes to cover every case, but I could just add a test that confirms a successful search for, e.g., an ISBN in `q`.

JPrevost (approved):

app/models/opensearch.rb
```diff
@@ -198,4 +199,42 @@ def aggregations
...
+      match: {
+        field => @params[field].downcase
+      }
+    }
+  end
+
+  def match_single_field_nested(field, match_array)
+    return unless @params[field]
+
+    match_array << {
```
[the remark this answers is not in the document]
jazairi: I think a fix for that for nested fields is to use `multi_match`. I've noted that in the rebased commit message. However, it seems like this is also the case for simple fields, and I _think_ `multi_match` only works for nested...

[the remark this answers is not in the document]
JPrevost: Yeah, I think it gets really complicated really fast and it may actually be how the fields are configured in the index so we have a bunch of directions to investigate once we get all the core bits built out.

PR #517, opened 2022-05-24
Add targeted field search to GraphQL V2

jazairi:

The data model specifies that certain fields must be
single-field searchable.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/RDI-103

This makes the following fields searchable in GraphQL:

* citation
* contributors
* funding_information
* identifiers
* location
* subjects
* title

This also renames the facet fields to avoid conflicts (see
side effects for more information on this).

_This introduces a breaking change_. There is some overlap
between the fields above and facets. In order to accommodate both
searching and faceting, '_facet' has been added to all of the
facets in the QueryType class, and the corresponding change has
been made in the Opensearch model.

This means that anyone using the previous syntax for facets will
see their queries fail. We will want to do more analysis of how
the API is used before deploying this to prod. If we see usage
of faceting in prod, we might consider renaming the search fields
instead of the facets.

NO

NO

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -46,6 +46,7 @@ module Types
         argument :from, String, required: false, default_value: '0'
 
         # applied facets
+        argument :collection_facet, [String], required: false, default_value: nil
         argument :content_type_facet, String, required: false, default_value: nil
         argument :contributors_facet, [String], required: false, default_value: nil
         argument :format_facet, [String], required: false, default_value: nil
@@ -118,6 +119,7 @@ module Types
         query[:location] = location
         query[:subjects] = subjects
         query[:title] = title
+        query[:collection_facet] = facets[:collection_facet]
         query[:content_format_facet] = facets[:format_facet]
         query[:content_type_facet] = facets[:content_type_facet]
         query[:contributors_facet] = facets[:contributors_facet]
@@ -137,7 +139,7 @@ module Types
         query[:language] = facets[:languages]
         query[:literary_form] = facets[:literary_form]
         query[:source] = facets[:source] if facets[:source] != 'All'
-        query[:subjects_facet] = facets[:subjects]
+        query[:subjects] = facets[:subjects]
         query
       end
     end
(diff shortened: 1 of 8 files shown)


app/graphql/types/query_type.rb
```diff
@@ -122,7 +137,7 @@ def construct_query(searchterm, facets)
         query[:language] = facets[:languages]
         query[:literary_form] = facets[:literary_form]
         query[:source] = facets[:source] if facets[:source] != 'All'
-        query[:subject] = facets[:subjects]
+        query[:subjects_facet] = facets[:subjects]
```
matt-bernhardt, line 140: I may have missed this discussion, but it looks like this change is being made to the v1 side of the flipflop? It doesn't look like this is matched by anything in the search model, so I'm not sure if I'm missing something.

app/models/opensearch.rb
```diff
@@ -88,19 +88,19 @@ def matches
   # https://www.elastic.co/guide/en/elasticsearch/reference/current/query-filter-context.html
   def filters
     f = []
-    f.push filter(@params[:collection], 'collections') if @params[:collection]
-    f.push filter(@params[:contributor], 'contributors') if @params[:contributor]
+    f.push filter(@params[:collection_facet], 'collections') if @params[:collection_facet]
```
matt-bernhardt, line 91: Is the `collection_facet` filter supposed to be addressed in the `query_type` file as well? I see the other seven fields, but not this one.

matt-bernhardt (commented):
I've got two questions about the lines I noted above, but in general I think this is solid.

My more general concern is about the timeline between merging this and getting it into production. It sounds like we still need to start collecting data about the use of these fields to confirm this approach, so it may not be a quick process? I worry about doing dependency updates - although maybe we can create a `v1` branch or something from prior to this merging, and use that in the meantime.

If the two questions above are no issue, and the timeline seems doable, then I'm fine with this merging as-is.

app/graphql/types/query_type.rb
JPrevost: Yeah I think you are correct. As this should still be using ElasticSearch on the v1 side of the if/else this should remain the old name. I think.

app/models/opensearch.rb
JPrevost: Yeah I think so. It's complicated because some of the facets don't actually work at all right now... but I think you are correct in that it won't show up as an option in GraphQL until the `query_type` file is updated.

JPrevost (approved):
@matt-bernhardt can you give the last commit I just added a look.

PR #521, opened 2022-06-24
Initial addition of opensearch highlights to graphql

JPrevost:

Since we show minimal metadata in the brief results, it might be
useful for end users to see where in the metadata their search terms
matched.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/RDI-156

This enables the OpenSearch feature to highlight query matches and
updates the GraphQL configuration to return that in the results.
Because we target `_source` as our GraphQL records, and the
highlights appear in `hits`, we inject the field into each record
in the QueryType class.

This approach also allows to return `score` in GraphQl results. Because
`score` will be useful to us in the near future, this commit configures
that as well.

* We will likely want to reevaluate how we do this after removing
the deprecated V1 fields. I've added a comment to remind of this to
QueryType#inject_hits_fields_into_source, but we should also open a
JIRA ticket once we confirm that this approach is acceptable in the
interim.
* Highlights are returned as an array of strings, so we'll need to
transform this to unserialized JSON in TIMDEX-UI.
* Some unrelated test assertions had to be updated, presumably
because more data sources had been loaded since the last time
cassettes were regenerated. If this becomes burdensome, we might
want to consider using staging OpenSearch and filtering secrets.

#### Developer

#### Code Reviewer

NO

NO


matt-bernhardt (approved):
This looks fine to me, and I agree with the strategy of making a ticket to streamline this implementation after the v1 implementation has been removed.

I asked one question about the `inject_hits_fields_into_sources` method, but only to understand a detail of the approach taken - I don't think it needs to change.

So long as Helen signs off on this from the data model perspective, I think this is okay from the code and testing side.

app/graphql/types/query_type.rb
```diff
@@ -89,10 +89,24 @@ def search(searchterm:, citation:, contributors:, funding_information:, identifi
...
         response
       end
+
+      # Long-term, we will probably want to define these fields discretely in RecordType. However, this might end up
+      # adding confusion while we are maintaining deprecated fields in that class. We should refactor this either soon
+      # after removing GraphQL V1 or as part of that work.
+      def inject_hits_fields_into_source(hits)
+        modded_source = []
+        hits.each do |hit|
+          source = hit['_source']
```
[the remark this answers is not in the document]
JPrevost: modded_source is an array of modified `source`s. It's possible `modded_sources` would be a better name to make that more clear.

app/graphql/types/record_type.rb
```diff
@@ -126,6 +126,19 @@ class LocationType < Types::BaseObject
       field :kind, String
       field :value, String
     end
+
+    class HighlightType < Types::BaseObject
+      field :kind, String
```
JPrevost, line 131: I wonder if `:field` or `:matched_field` would be more useful here than `:kind`?

app/graphql/types/record_type.rb
```diff
@@ -126,6 +126,19 @@ class LocationType < Types::BaseObject
...
+      field :value, [String]
```
JPrevost, line 132: Similarly, I wonder if `:matched_phrases` or something more descriptive than `:value` would be more useful...

hakbailey (approved):
Looks great!

PR #570, opened 2022-08-19
Upgrade to Rails 7

jazairi:

We want to stay up to date with the latest Rails.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ENGX-155

This upgrades to Rails 7 and loads the new framework defaults.
Prior to this, I ran `rails app:update`, checked the diffs for all
files, and confirmed that the new framework defaults would not
cause problems. In this case, it seemed like there were no Rails
files that needed to be updated, so I kept the existing files.

* The way we handle domain redirects
(`ApplicationController#ensure_domain`) now throws an unsafe
redirect error. I've added the `allow_other_host` param to that
method to bypass this. In this case the other host would come
from env, so the risk seems low, but I'd encourage the reviewer
to consider any edge cases I may have missed.
* The only new framework default that seemed like it could cause
problems for us was the key generator digest changing to SHA256:
https://guides.rubyonrails.org/v7.0/upgrading_ruby_on_rails.html#key-generator-digest-class-changing-to-use-sha256.
However, I don't think we are using encrypted cookies
in this app, so I don't anticipate problems there.

#### Developer

#### Code Reviewer

NO

YES


app/controllers/application_controller.rb
```diff
@@ -9,6 +9,6 @@ def ensure_domain
     return if request.host == ENV['PREFERRED_DOMAIN']
 
     Rails.logger.info("Handling Domain Redirect: #{request.host}")
-    redirect_to "https://#{ENV['PREFERRED_DOMAIN']}", status: :moved_permanently
+    redirect_to "https://#{ENV['PREFERRED_DOMAIN']}", status: :moved_permanently, allow_other_host: true
```
JPrevost, line 12: I'm comfortable with this since we control the redirect target as you noted. If we redirected to a URL supplied target it could be used to trick people to think they were on our domain much like ezproxy was targeted before it moved to a safelist model for redirects. Makes sense this now requires allowing explicitly.

jazairi:
Thanks, @JPrevost. I opened that ticket here: https://mitlibraries.atlassian.net/browse/TIMX-86

PR #642, opened 2023-01-26
Add descriptions to GraphQL V2 fields

jazairi:
Why these changes are being introduced:

Many of the GraphQL fields are not documented.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TIMX-116

How this addresses that need:

This adds descriptions to all V2 fields and search arguments, using the TIMDEX Data Model v2.0 document as a guide.

Side effects of this change:

* Descriptions have been modified from the data model where applicable.
* Listing record sources is a little unwieldy now that we have six of them, but I still think it's useful to include here. I'd be interested in ideas of how better to display this information.
* One field, 'numbering', does not have a description in the data model. I've set this aside until we have a better understanding of what it does.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -44,26 +44,37 @@ module Types
 
       field :search, SearchType, null: false,
                                  description: 'Search for timdex records' do
-        argument :searchterm, String, required: false, default_value: nil
-        argument :citation, String, required: false, default_value: nil
-        argument :contributors, String, required: false, default_value: nil
-        argument :funding_information, String, required: false, default_value: nil
-        argument :identifiers, String, required: false, default_value: nil
-        argument :locations, String, required: false, default_value: nil
-        argument :subjects, String, required: false, default_value: nil
-        argument :title, String, required: false, default_value: nil
-        argument :from, String, required: false, default_value: '0'
+        argument :searchterm, String, required: false, default_value: nil, description: 'Query all searchable fields'
+        argument :citation, String, required: false, default_value: nil, description: 'Search by citation information'
+        argument :contributors, String, required: false, default_value: nil,
+                                        description: 'Search by contributor name; e.g., author, editor, etc.'
+        argument :funding_information, String, required: false, default_value: nil,
+                                               description: 'Search by funding information; e.g., funding source, award name, etc.'
+        argument :identifiers, String, required: false, default_value: nil,
+                                       description: 'Search by unique indentifier; e.g., ISBN, DOI, etc.'
+        argument :locations, String, required: false, default_value: nil, description: 'Search by locations'
+        argument :subjects, String, required: false, default_value: nil, description: 'Search by subject terms'
+        argument :title, String, required: false, default_value: nil, description: 'Search by title'
+        argument :from, String, required: false, default_value: '0',
+                                description: 'Search result number to begin with (the first result is 0)'
         argument :index, String, required: false, default_value: nil,
                                  description: 'It is not recommended to provide an index value unless we have provided you with one for your specific use case'
 
         # applied facets
-        argument :content_type_facet, [String], required: false, default_value: nil
-        argument :contributors_facet, [String], required: false, default_value: nil
-        argument :format_facet, [String], required: false, default_value: nil
-        argument :languages_facet, [String], required: false, default_value: nil
-        argument :literary_form_facet, String, required: false, default_value: nil
-        argument :source_facet, [String], required: false, default_value: nil
-        argument :subjects_facet, [String], required: false, default_value: nil
+        argument :content_type_facet, [String], required: false, default_value: nil,
+                                                description: 'Filter results by content type. Use the `contentType` aggregation for a list of possible values'
+        argument :contributors_facet, [String], required: false, default_value: nil,
+                                                description: 'Filter results by contributor. Use the `contributors` aggregation for a list of possible values'
+        argument :format_facet, [String], required: false, default_value: nil,
+                                          description: 'Filter results by format. Use the `format` aggregation for a list of possible values'
+        argument :languages_facet, [String], required: false, default_value: nil,
+                                             description: 'Filter results by language. Use the `languages` aggregation for a list of possible values'
+        argument :literary_form_facet, String, required: false, default_value: nil,
+                                               description: 'Filter results by fiction or nonfiction'
+        argument :source_facet, [String], required: false, default_value: nil,
+                                          description: 'Filter by source record system. Use the `sources` aggregation for a list of possible values'
+        argument :subjects_facet, [String], required: false, default_value: nil,
+                                            description: 'Filter by subject terms. Use the `contentType` aggregation for a list of possible values'
       end
     else
       def record_id(id:)
--- a/app/graphql/types/record_type.rb
+++ b/app/graphql/types/record_type.rb
@@ -1,18 +1,18 @@
 module Types
   class LinkType < Types::BaseObject
-    field :kind, String, null: true
-    field :text, String, null: true
-    field :url, String, null: false
-    field :restrictions, String, null: true
+    field :kind, String, null: true, description: 'Type of link'
+    field :text, String, null: true, description: 'Additional description of a link, if applicable'
+    field :url, String, null: false, description: 'URL of a link'
+    field :restrictions, String, null: true, description: 'Restrictions on a link, if available'
   end
 
   if Flipflop.v2?
     class ContributorType < Types::BaseObject
-      field :kind, String, null: true
-      field :value, String, null: false
-      field :identifier, [String], null: true
-      field :affiliation, [String], null: true
-      field :mit_affiliated, Boolean, null: true
+      field :kind, String, null: true, description: 'Type of contributor; e.g., editor, author, etc.'
+      field :value, String, null: false, description: 'Name of contributor'
+      field :identifier, [String], null: true, description: 'Unique identifier(s) of a contributor'
+      field :affiliation, [String], null: true, description: 'Institutional affiliation(s) of a contributor'
+      field :mit_affiliated, Boolean, null: true, description: 'Identifies whether a contributor is affiliated with MIT'
     end
   else
     class ContributorType < Types::BaseObject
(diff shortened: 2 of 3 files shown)


app/graphql/types/query_type.rb
```diff
@@ -44,26 +44,35 @@ def record_id(id:, index:)
...
+        argument :contributors_facet, [String], required: false, default_value: nil,
+                                                description: 'Filter results by contributors'
+        argument :format_facet, [String], required: false, default_value: nil,
+                                          description: 'Filter results by format; e.g., "Print Volume", "DVD", etc.'
+        argument :languages_facet, [String], required: false, default_value: nil,
+                                             description: 'Filter results by languages'
+        argument :literary_form_facet, String, required: false, default_value: nil,
+                                               description: 'Filter results by fiction or nonfiction'
+        argument :source_facet, [String], required: false, default_value: nil,
+                                          description: "Filter by source record system:\n* MIT Alma\n* MIT ArchivesSpace\n* DSpace@MIT\n* Abdul Latif Jameel Poverty Action Lab Dataverse\n* Woods Hole Open Access Server\n* Zenodo"
```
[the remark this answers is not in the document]
JPrevost: Alternatively, we could document the available sources and what they include in our general documentation and once that exists provide a link here?

app/graphql/types/record_type.rb
```diff
@@ -146,48 +146,55 @@ def matched_phrases
...
+      field :edition, String, null: true, description: 'Edition information for item'
       field :imprint, [String], null: true, deprecation_reason: 'Use `publicationInformation`'
-      field :publication_information, [String]
-      field :physical_description, String, null: true
-      field :publication_frequency, [String], null: true
+      field :publication_information, [String], description: 'Imprint information for item'
+      field :physical_description, String, null: true, description: 'Physical description of item'
+      field :publication_frequency, [String], null: true,
+                                              description: 'Publication frequency of item (used for serials)'
       field :numbering, String, null: true
```
JPrevost, line 177: @hakbailey do you have suggestion as to what wording we might use to describe `numbering`? It wasn't defined in the data model documents used to add descriptions to the other fields.

app/graphql/types/query_type.rb
```diff
@@ -44,26 +44,35 @@ def record_id(id:, index:)
...
 
         # applied facets
-        argument :content_type_facet, [String], required: false, default_value: nil
-        argument :contributors_facet, [String], required: false, default_value: nil
-        argument :format_facet, [String], required: false, default_value: nil
-        argument :languages_facet, [String], required: false, default_value: nil
-        argument :literary_form_facet, String, required: false, default_value: nil
-        argument :source_facet, [String], required: false, default_value: nil
-        argument :subjects_facet, [String], required: false, default_value: nil
+        argument :content_type_facet, [String], required: false, default_value: nil,
```
JPrevost, line 63: I think I want to rename these arguments (outside the scope of this work but noting it here). We should very likely rename these to replace `_facet` with `_filter`. I'll either open a ticket or include with my deprecation evaluation work.

app/graphql/types/query_type.rb
```diff
@@ -44,26 +44,35 @@ def record_id(id:, index:)
...
-        argument :contributors, String, required: false, default_value: nil
-        argument :funding_information, String, required: false, default_value: nil
-        argument :identifiers, String, required: false, default_value: nil
-        argument :locations, String, required: false, default_value: nil
-        argument :subjects, String, required: false, default_value: nil
-        argument :title, String, required: false, default_value: nil
-        argument :from, String, required: false, default_value: '0'
+        argument :searchterm, String, required: false, default_value: nil, description: 'Query all searchable fields'
+        argument :citation, String, required: false, default_value: nil, description: 'Search by citation information'
+        argument :contributors, String, required: false, default_value: nil, description: 'Search by contributor names'
```
JPrevost, line 49: What do you think about something like `Search by contributor/author names` to make it clearer what contributor means?

app/graphql/types/query_type.rb
```diff
@@ -44,26 +44,35 @@ def record_id(id:, index:)
...
-        argument :citation, String, required: false, default_value: nil
-        argument :contributors, String, required: false, default_value: nil
-        argument :funding_information, String, required: false, default_value: nil
-        argument :identifiers, String, required: false, default_value: nil
-        argument :locations, String, required: false, default_value: nil
-        argument :subjects, String, required: false, default_value: nil
-        argument :title, String, required: false, default_value: nil
-        argument :from, String, required: false, default_value: '0'
+        argument :searchterm, String, required: false, default_value: nil, description: 'Query all searchable fields'
+        argument :citation, String, required: false, default_value: nil, description: 'Search by citation information'
```
JPrevost, line 48: I sort of want to have a way to note arguments as `we don't recommend this as 99.999% of our data doesn't include this`. <sigh>

app/graphql/types/query_type.rb
```diff
@@ -44,26 +44,35 @@ def record_id(id:, index:)
...
-        argument :literary_form_facet, String, required: false, default_value: nil
-        argument :source_facet, [String], required: false, default_value: nil
-        argument :subjects_facet, [String], required: false, default_value: nil
+        argument :content_type_facet, [String], required: false, default_value: nil,
+                                                description: 'Filter results by content type'
+        argument :contributors_facet, [String], required: false, default_value: nil,
+                                                description: 'Filter results by contributors'
+        argument :format_facet, [String], required: false, default_value: nil,
+                                          description: 'Filter results by format; e.g., "Print Volume", "DVD", etc.'
+        argument :languages_facet, [String], required: false, default_value: nil,
```
JPrevost, line 69: For all of these _facet fields, I wonder if there is a way to hint that the best way to get these values is from the corresponding aggregation. I don't have a specific suggestion on how to do that, so I'd be okay with just thinking about that as a future enhancement.

app/graphql/types/record_type.rb
```diff
@@ -146,48 +146,55 @@ def matched_phrases
...
-      field :timdex_record_id, ID, null: false
-      field :source, String, null: false
-      field :source_link, String, null: false
-      field :title, String, null: false
-      field :alternate_titles, [Types::AlternateTitleType], null: true
-      field :contributors, [Types::ContributorType], null: true
-      field :subjects, [Types::SubjectType], null: true
-      field :identifiers, [Types::IdentiferType], null: true
+      field :timdex_record_id, ID, null: false, description: 'TIMDEX unique identifier for the item'
+      field :source, String, null: false,
```
JPrevost, line 150: This formatting doesn't come across as intended in our generated Reference Docs. I think a comma separated list without the line breaks might works best here.

app/graphql/types/record_type.rb
```diff
@@ -146,48 +146,55 @@ def matched_phrases
...
+                             description: 'Name of source record system. One of the following:\n* MIT Alma\n* MIT ArchivesSpace\n* DSpace@MIT\n* Abdul Latif Jameel Poverty Action Lab Dataverse\n* Woods Hole Open Access Server\n* Zenodo'
+      field :source_link, String, null: false, description: 'URL for source record in source system'
+      field :title, String, null: false, description: 'Title of item'
+      field :alternate_titles, [Types::AlternateTitleType], null: true, description: 'Alternate titles for the item'
+      field :contributors, [Types::ContributorType], null: true, description: 'Contributors to the item'
```
JPrevost, line 155: `Contributors/Authors to the item`?

JPrevost (commented):
I think we need to drill down and document the additional nested fields and aggregations, but I'd be okay with doing that as a second pass and working to land this initial round first.

None of my comments are blocking.

app/graphql/types/query_type.rb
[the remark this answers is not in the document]
jazairi: Yeah, I like the idea of linking to general docs. I'll remove the list here and open a ticket to come back to that.

app/graphql/types/query_type.rb
jazairi: Yeah, `_filter` seems more accurate to me, since they're applied before the search is executed rather than refining the results.

app/graphql/types/query_type.rb
jazairi: For sure. I might tweak it a bit further to provide a couple contributor examples, like "e.g., author names, editor names, etc."

app/graphql/types/query_type.rb
jazairi: I hear you. I think tutorial/explanation docs would be a good place to explain some of the limitations.

app/graphql/types/query_type.rb
jazairi: Might be good to include in how-to/reference docs? That said, I guess these very docs could fall into either of those categories. I'm not opposed to adding a note to that effect to each of the facet fields.

JPrevost: Good point. I like the idea of both including an example of how this works in our docs and linking to that example from here once written.

jazairi: Lol yeah, after commenting I realized I could cut down the word count substantially. English major tendencies. :)

jazairi:
@JPrevost I just pushed another commit to address your feedback and document subfields and aggregations. Please let me know if I missed anything, and if any of the subfield descriptions can be improved (some are better than others).

app/graphql/types/record_type.rb
```diff
@@ -1,18 +1,18 @@
 module Types
   class LinkType < Types::BaseObject
-    field :kind, String, null: true
-    field :text, String, null: true
-    field :url, String, null: false
-    field :restrictions, String, null: true
+    field :kind, String, null: true, description: 'Type of link'
+    field :text, String, null: true, description: 'Additional description of a link, if applicable'
+    field :url, String, null: false, description: 'URL of a link'
+    field :restrictions, String, null: true, description: 'Restrictions on a link, if applicable'
```
JPrevost, line 6: For this one, I think I'd prefer `if available`. We likely have restrictions on lots of links but it isn't in the metadata.

app/graphql/types/query_type.rb
```diff
@@ -61,18 +62,19 @@ def record_id(id:, index:)
 
         # applied facets
         argument :content_type_facet, [String], required: false, default_value: nil,
-                                                description: 'Filter results by content type'
+                                                description: 'Filter results by content type. Use the `contentType` aggregation of a list of possible values'
```
JPrevost, line 65: I like these additions, but this set I think we have to many `of`s and not enough `for`s? (I think either `of` could be replaced with a `for` maybe? Applies to all in this class of description.

JPrevost (commented):
Couple more small suggested changes. I'm super excited to have these descriptions in the spec. Thanks!

app/graphql/types/query_type.rb
jazairi: Serves me right for pushing code after 5pm. 🤦

JPrevost (approved):
Once this merges go ahead and run the `Deploy Jekyll with GitHub Pages dependencies preinstalled` action and choose the `main` branch so we can see the new ref docs build with these new descriptions (and so you can see how that process works first hand).

PR #643, opened 2023-01-31
Reintroduce deprecated source filter 

JPrevost:

* The change we are making to source filtering is a breaking change that affects an existing system. Deprecating to allow for a smoother launch rather than requiring launching both at exactly the same time will make this a lot easier to isolate the changes.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TIMX-178

* reintroduces deprecated `source` argument
* changes deprecated `identifier` back to `ID` to match legacy type to match the legacy spec we are deprecating
  * regenerates cassette for current sourceFacet (see side effects)

* In regenerating the cassette for sourceFacet I noticed we don't have an easy way to genereate cassettes from our AWS Opensearch without leaving remnants in the cassettes we likely don't want (i.e. we don't scrub well). We should add some scrubbers to replace the potentially sensitive headers with fake data as well as replace the uri with a fake uri (although inserting `uri: http://localhost:9200/...` as our fake uri would work as well). Once we make this process easier to use our AWS based data in these tests we should regenerate all of the cassettes and update our documentation to prefer using real data to genereate the cassettes to make it easier to maintain

#### Developer

#### Code Reviewer

NO

NO


JPrevost:
The following two queries should result in the same results in this PR build.

```graphql
{
  search(searchterm: "lynch", sourceFacet:["MIT ArchivesSpace"]) {
    hits
    records {
      title
      timdexRecordId
    }
  }
}
```

```graphql
{
  search(searchterm: "lynch", source: "MIT ArchivesSpace") {
    hits
    records {
      title
      timdexRecordId
    }
  }
}
```

jazairi (approved):
Good call re: the need for better scrubbing of our cassettes. Could you open a ticket for that if you haven't already?

JPrevost:
Good call on making sure we open a ticket before moving on. It's sparse, but https://mitlibraries.atlassian.net/browse/TIMX-179

jazairi:
@JPrevost Thanks! I think that background is all we need in there for now.

PR #644, opened 2023-02-01
Conditional highlight requests

JPrevost:

* Highlights seem to introduce additional time to some opensearch queries so only requesting them when we want them makes sense

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TIMX-173

* adds a GraphQL Tracer to expose our Analyzer data to be available before we make the opensearch call
* Updates our analyzer to return data instead of just logging (it still logs as well)
* provides a new boolean arugment to OpenSearch.new.search to indicate if `highlights` are needed
* Updates OpenSearch build_query to only include highlights if requested

* We should confirm timdex-search-ui continues to work without changes
   - [x] Confirmed timdex-ui continues to work with this change with no changes (aka highlights always return)
* We should add a UI element to timdex-search-ui to only return highlights if selected
  - [x] [Jira ticket for this work](https://mitlibraries.atlassian.net/browse/TIMX-182)

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/opensearch.rb
+++ b/app/models/opensearch.rb
@@ -2,8 +2,9 @@ class Opensearch
   SIZE = 20
   MAX_PAGE = 200
 
-  def search(from, params, client, index = nil)
+  def search(from, params, client, highlight = false, index = nil)
     @params = params
+    @highlight = highlight
     index = default_index unless index.present?
     client.search(index: index,
                   body: build_query(from))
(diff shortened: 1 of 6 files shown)


app/models/opensearch.rb
```diff
@@ -2,8 +2,9 @@ class Opensearch
   SIZE = 20
   MAX_PAGE = 200
 
-  def search(from, params, client, index = nil)
+  def search(from, params, client, highlight = false, index = nil)
```
JPrevost, line 5: Rubocop notes that for boolean parameters we should consider keyword arguments. I suspect that's worth considering but felt like moving `index` to a keyword would make sense and while simple in the code would require a lot of test updates which I didn't want to introduce in this changeset. The suggestion is logged in code climate as technical debt and rather than _also_ opening a Jira ticket I thought we might introduce a practice in EngX of actually looking back on our code climate logged debt more frequently to look for maintenance work. Thoughts?

jazairi: Yeah, that's an interesting idea. I feel like if we were to open Jira tickets for every piece of technical debt, our backlogs would get flooded.

jazairi (approved):
Interesting stuff. Thanks for taking this on!

PR #647, opened 2023-02-06
Playground

JPrevost:
The first commit replaces our old playground.html file with the latest example from the GraphiQL repository entirely.

The second adds a small change to which endpoint is loaded.

* Rather than our playground always pointing at a specific version of
  TIMDEX, this has the playground use the "local" version, which can
  include local development, PR builds, stage, prod, etc.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TIMX-183

* Updates to the latest version of the GraphiQL playground example
  code
* Adds support for using the running, or "local", version of the GraphQL
  endpoint

* There was an attempt to create a script to auto download the various
  3rd Party CDN resources and serve them localy. Most of the files
  worked fine, but `graphiql.min.js` consistently downloaded in a state
  in which whitespace was replaced with non-whitespace characters. It
  felt like an encoding issue, but it was unclear how to resolve it.
  Loading the file in a browser as text and copy/pasting worked, but
  downloading directly in a browser, using curl, wget, or http CLI all
  led to the same behavior. As such, relying on the 3rd Party CDN at
  this time seemed prudent.

#### Developer

#### Code Reviewer

NO

YES

### DIFF
--- a/public/playground.html
+++ b/public/playground.html
@@ -1,59 +1,176 @@
+<!--
+ *  Copyright (c) 2021 GraphQL Contributors
+ *  All rights reserved.
+ *
+ *  MIT License
+ * 
+ *  Permission is hereby granted, free of charge, to any person obtaining a copy
+ *  of this software and associated documentation files (the "Software"), to deal
+ *  in the Software without restriction, including without limitation the rights
+ *  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


jazairi (approved):
While I understand your concern about relying on the CDN, it feels okay to me in this case since the publisher is trustworthy(ish). :shipit:

PR #665, opened 2023-02-27
Fix default index for V2 retrieve queries

jazairi:
Why these changes are being introduced:

The Retrieve model uses the default elasticsearch index name for V1 and V2 queries, which means that V2 retrieve queries use the wrong index name.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TIMX-185

How this addresses that need:

This updates Retrieve#default_index to use the default opensearch index name for V2 queries.

Side effects of this change:

The regression tests for this pass, but somewhat disingenously as we currently use the same index names in .env.test for elasticsearch and opensearch. We should probably regenerate cassettes with the latest prod data and start using the `all-current` index for V2 queries.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/test/models/retrieve_test.rb
+++ b/test/models/retrieve_test.rb
@@ -1,0 +1,17 @@
+require 'test_helper'
+
+class RetrieveTest < ActiveSupport::TestCase
+  def setup
+    @test_strategy = Flipflop::FeatureSet.current.test!
+  end
+
+  test 'graphlv1 uses correct index for retrieve queries' do
+    @test_strategy.switch!(:v2, false)
+    assert_equal ENV.fetch('ELASTICSEARCH_INDEX'), Retrieve.new.default_index
+  end
+
+  test 'graphqlv2 uses correct index for retrieve queries' do
+    @test_strategy.switch!(:v2, true)
+    assert_equal ENV.fetch('OPENSEARCH_INDEX'), Retrieve.new.default_index
+  end
+end
--- a/app/models/retrieve.rb
+++ b/app/models/retrieve.rb
@@ -16,7 +16,7 @@ class Retrieve
   end
 
   def default_index
-    ENV.fetch('ELASTICSEARCH_INDEX', nil)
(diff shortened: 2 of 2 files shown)


JPrevost (approved):

test/models/retrieve_test.rb
```diff
@@ -1,13 +1,17 @@
 require 'test_helper'
 
 class RetrieveTest < ActiveSupport::TestCase
-  test 'uses correct index for V1 queries' do
-    ClimateControl.modify(V2: nil) do
-      assert_equal ENV.fetch('ELASTICSEARCH_INDEX'), Retrieve.new.default_index
-    end
+  def setup
+    @test_strategy = Flipflop::FeatureSet.current.test!
```
JPrevost, line 5: Interesting. I wonder if this would have allowed us to avoid the annoying excludev1/v2 split for the past several months  :)

jazairi:
@JPrevost I remembered the issue with the tests. `ClimateControl.modify(ENV_VAR: nil)` doesn't do what we expect it to. Theoretically, we can do [something like this](https://github.com/thoughtbot/climate_control#thread-safety), but I recall that approach is also kind of flakey. I just toggled the Flipflop feature instead and confirmed that that works with different values of `ELASTICSEARCH_INDEX` and `OPENSEARCH_INDEX` (though, as we discussed, this will not be necessary when we remove V1).

JPrevost (approved):
I've confirmed the default index works for opensearch now but does not in the prod v2 instance. I've confirmed we can still set the index specifically (which works in prod v2 as well).

I did not test v1. We can do that in the staging tier by setting one of the many stage instances delete the v2 env for a few minutes for testing.

test/models/retrieve_test.rb
jazairi: I got the idea from the graphql controller tests, so I'm guessing the weirdness persists either way. 🤷

PR #666, opened 2023-03-02
Add initial how-to guides

jazairi:

Version 2 of the API introduces significant changes that should be documented publicly for our end users.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TIMX-192

This adds an initial set of how-to guides, which provide specific example queries of certain features.

* There are some WAVE errors, primarily related to low contrast.
We might want to use the light color scheme when we launch.
* Since this is not considered an external link like the
SpectaQL-generated reference docs, it seems to always appear in
the nav bar above reference, no matter how the nav order is set.
This is probably fine, but if I were developing an app with TIMDEX,
the reference docs are probably what I would look up most frequently
(although maybe not first).

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/docs/guides/highlighting_search_terms.md
+++ b/docs/guides/highlighting_search_terms.md
@@ -1,0 +1,60 @@
+---
+layout: home
+title: Highlighting search terms
+parent: How-to guides
+nav_order: 7
+---
+
+## Highlighting search terms in results
+
+TIMDEX can return a list of search term matches in the results. The field that returns these matches is called
+'highlight':
+
+```graphql
+{
+  search(searchterm: "data") {
+    records {
+      highlight {
+        matchedField
+        matchedPhrases
+      }
+    }
+  }
+}
+```
+
+In the query above will return a list of matches for the search term 'data'. The 'highlight' field is nested; subfield
+'matchedField' will show the field that matched your search term, and subfield 'matchedPhrases' will show where in the
+field the search term was matched.
+
+Here's an example of a result:
+
+```json
+{
+  "highlight": [
+    {
+      "matchedField": "citation",
+      "matchedPhrases": [
+        "<span class=\"highlight\">Data</span> security and <span class=\"highlight\">data</span> processing. 1975."
+      ]
+    },
+    {
+      "matchedField": "title",
+      "matchedPhrases": [
+        "<span class=\"highlight\">Data</span> security and <span class=\"highlight\">data</span> processing"
+      ]
+    }
+  ]
+}
+```
+
+In this case, we found matches in the 'citation' and 'title' fields. The HTML included in the 'matchedPhrases' subfield
+is for display purposes. If you plan to display the TIMDEX data in a user interface, you can write a CSS rule on the
+'highlight' class to emphasize the text however you choose.
+
+While the example above shows the entire fields, 'matchedPhrases' has a character limit. If the field contains more
+than 100 characters, TIMDEX will return a 100-character fragment that contains the highlighted term.
+
+## See also
+
+- [Returning specific fields](returning_fields)
--- a/docs/guides/retrieving_records.md
+++ b/docs/guides/retrieving_records.md
@@ -1,0 +1,55 @@
+---
+layout: home
+title: Retrieving a single record
+parent: How-to guides
+nav_order: 3
+---
+
+## Retrieving a single record
+
+
+In addition to searching, it is also possible to retrieve a specific record. To do so, you will need the
+`timdexRecordId` for the item. You can find this by returning that field in a query, like so:
+
+```graphql
+{
+  search(searchterm: "pragmatic programmer") {
+    records {
+      title
+      timdexRecordId
+    }
+  }
+}
+
+```
+
+[Run this query in the GraphQL playground.](https://timdex.mit.edu/playground?query=%7B%0A%20%20search(searchterm%3A%20%22pragmatic%20programmer%22)%20%7B%0A%20%20%20%20records%20%7B%0A%20%20%20%20%20%20title%0A%20%20%20%20%20%20timdexRecordId%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A)
+
+Once you have the `timdexRecordId` of the record in question, you can retrieve information about that record only:
+
+```graphql
+{
+  recordId(id: "alma:9935059777706761") {
+    title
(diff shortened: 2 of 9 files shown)


docs/guides/highlighting_search_terms.md
```diff
@@ -0,0 +1,60 @@
...
+    records {
+      highlight {
+        matchedField
+        matchedPhrases
+      }
+    }
+  }
+}
+```
+
```
JPrevost, line 25: What are your thoughts on having a "[Run this query in the Playground](/playground?query=%7B%0A%20%20%20search(searchterm%3A%20%22data%22)%20%7B%0A%20%20%20%20%20records%20%7B%0A%20%20%20%20%20%20%20highlight%20%7B%0A%20%20%20%20%20%20%20%20%20matchedField%0A%20%20%20%20%20%20%20%20%20matchedPhrases%0A%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%7D%0A%20%20%20%7D%0A%20%7D)" link for each example?

I generally think it will be good to do that but would be okay with coming back to that later as well.

docs/guides/using_filters.md
```diff
@@ -0,0 +1,33 @@
...
+---
+
+## Using filters to narrow your search
+
+TIMDEX includes a set of filters that you can use to limit your search. The [API reference](https://mitlibraries.github.io/timdex/reference/#query-search)
+for the search operation lists each available filter and how to use them. In this how-guide, we'll look at the
+sourceFilter, but the process is similar with other filters. If you're not sure what values a filter will accept, try
+a search that [returns aggregations](viewing_aggregations).
+
+This example demonstrates a query that searches for the term "data" in the two sources dspace@mit and zenodo. We are
```
JPrevost, line 15: Let's use "not Zenodo" in this example. dspace and alma or dspace and aspace as I don't recall if we'll leave zenodo in as a default search (I think it's supposed to require a separate index but we may not have configured our endpoints to do that yet...)

docs/guides/retrieving_records.md
```diff
@@ -0,0 +1,50 @@
...
+      location
+    }
+    identifiers {
+      kind
+      value
+    }
+  }
+}
+```
+
```
JPrevost, line 44: Maybe include something like:

"You can instead [request all of the fields](returning_fields) you want in the original search query instead of doing this multistep process if it makes more sense for your use case." ?

JPrevost (commented):
I've made a few comments and agree for now let's switch the default to light mode as I'm not sure how to override css in our current build process (it was easy in the old repo so I'm sure we can bring it back to this as well but for now the contrast issues are solved by changing to light mode as you suggested).

docs/guides/viewing_aggregations.md
```diff
@@ -0,0 +1,38 @@
...
+    source {
+      docCount
+      key
+    }
+  }
+}
+```
+
+Note that all aggregations are multivalue fields that require the `docCount` and `key` subfields. If you are using the
+[GraphQL playground](https://timdex.mit.edu/playground) and do not include these, it should add them automatically.
```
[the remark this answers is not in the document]
jazairi: It added them for me yesterday, but I guess it may have auto-suggested them and I provided some input to accept the suggestion before I even saw it.

I share your waffliness on this one. Since we're both feeling iffy about it, I'd suggest removing it.

docs/guides/highlighting_search_terms.md
jazairi: On second thought, there's nothing stopping us from using `timdex.mit.edu`. The queries just won't work until we launch, which I think is fine.

docs/guides/viewing_aggregations.md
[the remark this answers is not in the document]
JPrevost: Oh it totally worked with no input and it's cool, but yeah let's pull that tip out for now until we can find a way to integrate aside "tips" or something in the markdown.

docs/guides/highlighting_search_terms.md
JPrevost: Good point. On launch we'll link prominently to these docs. Before that they are hidden behind having to know the link which I suspect most people aren't interested enough to dig up.

docs/guides/viewing_aggregations.md
```diff
@@ -0,0 +1,38 @@
+---
+layout: home
+title: Viewing aggregations
+parent: How-to guides
+nav_order: 6
+---
+
+## Viewing search aggregations
+Aggregations list the total count of search results meeting a certain criterion. For example, you may want to know how
```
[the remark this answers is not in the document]
jazairi: Good, that's my preference. :) Missed this one, though. Thanks.

docs/guides/retrieving_records.md
jazairi: Are all of those fields also available in `search`?

JPrevost: Yup. Every field is available to be requested in Search.

docs/guides/using_filters.md
```diff
@@ -0,0 +1,35 @@
...
+---
+
+## Using filters to narrow your search
+
+TIMDEX includes a set of filters that you can use to limit your search. The [API reference](https://mitlibraries.github.io/timdex/reference/#query-search)
+for the search operation lists each available filter and how to use them. In this how-guide, we'll look at the
+sourceFilter, but the process is similar with other filters. If you're not sure what values a filter will accept, try
+a search that [returns aggregations](viewing_aggregations).
+
+This example demonstrates a query that searches for the term "data" in the "dspace@mit" or the "abdul latif jameel
```
JPrevost, line 15: We aren't maintaining daily updates at all for any of the RDI sources. The only sources we will be promoting as available are: dspace, aspace, alma/catalog. I'll share more info on this in Slack to make sure we're on the same page about this as I suspect both that we haven't communicated this well _and_ we may actually need to confirm what the APIs do by default matches our intent.

JPrevost (changes requested):
See note about the sources in the example. I'll check in on Slack to clarify and talk about what we need to tweak beyond these docs to reflect this, but for now please just use our 3 core sources in any examples.`mit alma`, `dspace@mit`, `mit archivesspace`.

JPrevost (approved):

PR #673, opened 2023-03-14
Explicitly define which fields to search

JPrevost:

* We ran into a `too_many_nested_clauses` error in opensearch for some queries
* Making this change seems to fix the issue

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TIMX-210

* I've gone through our [opensearch mappings](https://github.com/MITLibraries/timdex-index-manager/blob/main/config/opensearch_mappings.json) and identified the fields that look like we expected to be searched during a routine search and added them to an array that will be searched.

* This avoids the problem and is likely find, but it isn't entirely clear why this was a problem or why this avoids the problem.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/opensearch.rb
+++ b/app/models/opensearch.rb
@@ -94,7 +94,10 @@ class Opensearch
     if @params[:q].present?
       m << {
         multi_match: {
-          query: @params[:q].downcase
+          query: @params[:q].downcase,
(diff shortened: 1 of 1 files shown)


jazairi (approved):
I'm not going to pretend I understand _why_ this fixes the problem, but the fact that it does is good enough for me. (It also feels better than just upping the `maxClauseCount` in OS.) :shipit:

PR #781, opened 2024-02-05
Adds geo_distance query to opensearch

JPrevost:
This has not yet been integrated with GraphQL, so the best way to validate this work is to use the rails console.

```ruby
bin/rails console
q = {geopoint:{distance: "1mi", latitude: "45", longitude: "-112"}}
Opensearch.new.search(0, q, Timdex::OSClient)
```

You can also combine keyword (or filters) using our existing query syntax:

```
q = {geopoint:{distance: "1mi", latitude: "45", longitude: "-112"}, searchterm: "train stations"}
```

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/GDT-135

* Adds an optional geo_distance search to our existing opensearch query builder

* rubocop config has been updated to the current version of ruby we are using
* many opensearch tests were updated as the previous versions continued to pass even if I changed the expected values

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/opensearch.rb
+++ b/app/models/opensearch.rb
@@ -6,7 +6,7 @@ class Opensearch
     @params = params
     @highlight = highlight
     index = default_index unless index.present?
-    client.search(index: index,
+    client.search(index:,
                   body: build_query(from))
   end
 
@@ -17,10 +17,10 @@ class Opensearch
   # Construct the json query to send to elasticsearch
   def build_query(from)
     query_hash = {
-      from: from,
+      from:,
       size: SIZE,
-      query: query,
-      aggregations: aggregations
+      query:,
+      aggregations:
     }
 
     query_hash[:highlight] = highlight if @highlight
@@ -109,7 +109,28 @@ class Opensearch
     match_single_field_nested(:identifiers, m)
     match_single_field_nested(:locations, m)
     match_single_field_nested(:subjects, m)
+
+    match_geodistance(m) if @params[:geodistance].present?
     m
+  end
+
+  def match_geodistance(match_array)
+    match_array << {
+      bool: {
+        must: {
+          match_all: {}
+        },
+        filter: {
+          geo_distance: {
+            distance: @params[:geodistance][:distance],
+            'locations.geoshape': {
+              lat: @params[:geodistance][:latitude],
+              lon: @params[:geodistance][:longitude]
+            }
+          }
+        }
+      }
+    }
   end
 
   # https://www.elastic.co/guide/en/elasticsearch/reference/current/query-filter-context.html
@@ -177,7 +198,7 @@ class Opensearch
     param.each do |source|
       sources << {
         term: {
-          source: source
+          source:
         }
       }
     end
--- a/test/models/opensearch_test.rb
+++ b/test/models/opensearch_test.rb
@@ -4,43 +4,50 @@ class OpensearchTest < ActiveSupport::TestCase
   test 'matches citation' do
     os = Opensearch.new
     os.instance_variable_set(:@params, { citation: 'foo' })
-    assert os.matches.select { |m| m[:citation] == 'foo' }
+
+    assert(os.matches.to_json.include?('{"match":{"citation":"foo"}}'))
   end
 
   test 'matches title' do
     os = Opensearch.new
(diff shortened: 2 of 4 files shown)


JPrevost:
FYI: code climate is being weird. Running the latest rubocop locally does not show the same errors as in code climate. They seem to be behind in versions (again) in a way that feels hard to work around. I'd like to propose we address that separately from this PR though.

JPrevost:
Latest commit seems to have rubocop being less weird... although now it's complaining about things not part of this PR. I'm hoping it will stabilize-ish after this merges maybe?

app/models/opensearch.rb
```diff
@@ -109,9 +109,30 @@ def matches
     match_single_field_nested(:identifiers, m)
     match_single_field_nested(:locations, m)
     match_single_field_nested(:subjects, m)
+
+    match_geopoint(m) if @params[:geopoint].present?
```
JPrevost, line 113: @matt-bernhardt did we agree on `geodistance` for this parameter? If so I can update this. I believe my code is from before we met to discuss naming for this feature.

app/models/opensearch.rb
```diff
@@ -17,10 +17,10 @@ def default_index
   # Construct the json query to send to elasticsearch
   def build_query(from)
     query_hash = {
-      from: from,
+      from:,
       size: SIZE,
-      query: query,
-      aggregations: aggregations
+      query:,
+      aggregations:
```
matt-bernhardt, line 23: I suspect that this is a syntactic sugar in Rails that I've forgotten - but the change in this hash is because the `query` and `aggregations` methods are called implicitly, and `from` is populated auto-magically?

app/models/opensearch.rb
JPrevost: I think I want to make something work before we bring them in on naming. I think since OpenSearch calls this a geodistance search we had agreed to that and I just hadn't updated the code. I'll make this `geodistance` as the parameter name we agree to for now and once implemented fully in graphql we can get feedback. Thanks!

app/models/opensearch.rb
```diff
@@ -177,7 +198,7 @@ def source_array(param)
     param.each do |source|
       sources << {
         term: {
-          source: source
+          source:
```
matt-bernhardt, line 201: Same question here as above - this is now being called implicitly?

test/models/opensearch_test.rb
```diff
@@ -4,59 +4,66 @@ class OpensearchTest < ActiveSupport::TestCase
   test 'matches citation' do
     os = Opensearch.new
     os.instance_variable_set(:@params, { citation: 'foo' })
-    assert os.matches.select { |m| m[:citation] == 'foo' }
+
+    assert(os.matches.to_json.include?('{"match":{"citation":"foo"}}'))
```
matt-bernhardt, line 8: I've confirmed that changing something in this test causes it to fail, but not that it was broken before. I'm happy to know that this test syntax is now fail-able, though.

matt-bernhardt (approved):
I have a question about what seems like a piece of syntactic sugar that we're relying on now, just to confirm that this is what's happening - but I don't think this is a blocking concern.

The feature itself being introduced seems to function as intended - I've run a few different queries in the console, at different locations and different distances, and it seems like it works as intended.

As long as I'm right about the syntactic sugar, then I'm :shipit:

app/models/opensearch.rb
JPrevost: Correct. When I changed the target ruby version in rubocop, this changed automatically when I save the file and seems to be the new preferred way of handling this. If we find we don't like it, we can decide to ignore the new rule. Other than being surprised it now works like this, I think I can learn to appreciate it.

app/models/opensearch.rb
JPrevost: Yup

JPrevost:
Thanks for the review @matt-bernhardt . I'm going to change the name from `geopoint` to `geodistance` as a new commit now

JPrevost:
@matt-bernhardt I've made the name changes if you would mind confirming the final commit makes sense to you before I squashy squash. Thanks!

matt-bernhardt (approved):
Yep, that looks good to me. I've run the new search name in the console and saw results come back.

Thanks!

PR #782, opened 2024-02-05
Updates format to be a single string

JPrevost:

* This was a bug in the graphql spec. TIMDEX has always treated format as a single string

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/GDT-151

* Updates GraphQL spec to correct the expectation to be a single string

* None of the TIMDEX sources mapped to anything other than a single string for `format`. Therefore, anyone requesting `format` via GraphQL would be getting an error in production at this time. This resolves that problem, so a deprecation does not feel appropriate. This was a bug that caused the inability to use a field, not a change in how a previously usable field is being mapped.
* Even though the previous comment remains true, we'll still need to refresh the spec in our deployed timdex ui instances as the spec has indeed changed.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/graphql/types/record_type.rb
+++ b/app/graphql/types/record_type.rb
@@ -163,7 +163,7 @@ module Types
     field :contents, [String], null: true, description: 'Table of contents for item'
     field :summary, [String], null: true,
                               description: 'Summary of contents of item (also where abstract goes if applicable)'
-    field :format, [String], null: true, description: 'Format of item e.g. "Print Volume", "DVD", etc.'
+    field :format, String, null: true, description: 'Format of item e.g. "Print Volume", "DVD", etc.'
     field :literary_form, String, null: true, description: 'Identifies the item as fiction or nonfiction'
     field :related_place, [String], null: true, deprecation_reason: 'Use `locations`'
     field :in_bibliography, [String], null: true, deprecation_reason: 'Use `related_items`'


jazairi:
I'm not getting any errors requesting `format` when testing this locally, but it's returning `null` for that field for seemingly every record. Do you have a sample search I can use to verify this that will return a record with a `format`?

ghukill:
> I'm not getting any errors requesting `format` when testing this locally, but it's returning `null` for that field for seemingly every record. Do you have a sample search I can use to verify this that will return a record with a `format`?

I don't know if helpful, but in Dev1 the record `gismit:EG_CAIRO_A25TOPO_1972` should have a value.  [This graphql query for this record](https://timdex-api-pipeline-dev.herokuapp.com/playground?query=query%20geoIndex%20%7B%0A%20%20recordId(id%3A%20%22gismit%3AEG_CAIRO_A25TOPO_1972%22)%20%7B%0A%20%20%20%20timdexRecordId%0A%20%20%20%20source%0A%20%20%20%20title%0A%20%20%20%20summary%0A%20%20%20%20sourceLink%0A%20%20%20%20links%20%7B%0A%20%20%20%20%20%20kind%0A%20%20%20%20%20%20text%0A%20%20%20%20%20%20url%0A%20%20%20%20%7D%0A%20%20%20%20locations%20%7B%0A%20%20%20%20%20%20value%0A%20%20%20%20%20%20kind%0A%20%20%20%20%7D%0A%20%20%20%20dates%20%7B%0A%20%20%20%20%20%20kind%0A%20%20%20%20%20%20value%0A%20%20%20%20%20%20range%20%7B%0A%20%20%20%20%20%20%20%20gte%0A%20%20%20%20%20%20%20%20lte%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20note%0A%20%20%20%20%7D%0A%20%20%20%20citation%0A%20%20%20%20contributors%20%7B%0A%20%20%20%20%20%20affiliation%0A%20%20%20%20%20%20identifier%0A%20%20%20%20%20%20kind%0A%20%20%20%20%20%20value%0A%20%20%20%20%7D%0A%20%20%20%20format%20%23%20this%20is%20the%20failing%20field%0A%20%20%20%20identifiers%20%7B%0A%20%20%20%20%20%20kind%0A%20%20%20%20%20%20value%0A%20%20%20%20%7D%0A%20%20%20%20languages%0A%20%20%20%20publicationInformation%0A%20%20%20%20subjects%20%7B%0A%20%20%20%20%20%20kind%0A%20%20%20%20%20%20value%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D&operationName=geoIndex) in Dev1 currently throws a 500.  

I believe the ticket had [another link](https://mitlibraries.atlassian.net/browse/GDT-151?focusedCommentId=121625) from @JPrevost that showed this works now, but I'm a little fuzzy on which Heroku URLs are pointing at which Opensearch instances (and it looks like that link no longer works).

ghukill (approved):
I read the change as going from an array of strings to a single string, which makes sense to me given that the [TIMDEX data model in Transmogrifier for the `format` field](https://github.com/MITLibraries/transmogrifier/blob/main/transmogrifier/models.py#L98) will always write a single string or nothing.

jazairi (approved):

PR #784, opened 2024-02-06
Update GraphQL to better support Locations

matt-bernhardt:
This replaces the GeoPoint field in LocationType with a GeoShape. I also take a crack at writing a test for this, but I'm honestly still not sure it is needed. I don't think we have any similar tests like this for the other types defined in `record_type.rb`. 

#### Developer

#### Code Reviewer

NO

NO


app/graphql/types/record_type.rb
```diff
@@ -108,7 +108,7 @@ class AlternateTitleType < Types::BaseObject
 
   # Warning: related_place was supposed to be an array but was incorrectly a string in graphql for v1
   class LocationType < Types::BaseObject
-    field :geopoint, String, description: 'GeoPoint data for the location, if applicable'
```
JPrevost, line 111: As we discussed on zoom, let's try to deprecate `geopoint` as a matter of best practice even though it is very unlikely anyone is using it.

So the we'd mark the field as deprecated and suggest using the new field while still mapping the new field value to the old field so people still get what they asked for either way. See https://github.com/MITLibraries/timdex/blob/main/app/graphql/types/record_type.rb#L34-L36 for an example of how we have done that on other fields that got renamed.

JPrevost (changes requested):

app/graphql/types/record_type.rb
matt-bernhardt: Sure thing - I've switched from a replacement to a deprecation for `geopoint` now. The test has been updated to include it as well. I see the deprecation announcement in the help docs, but don't see it in the actual API response, so I don't yet have a way to test for the deprecation itself.

JPrevost: I don't think we need an explicit test for the deprecation at this point. Manually confirm it is flagged as deprecated along with the tests that both fields return the expected data should be solid. I'll finish up the review on this now thanks!

app/graphql/types/record_type.rb
```diff
@@ -108,9 +108,14 @@ class AlternateTitleType < Types::BaseObject
 
   # Warning: related_place was supposed to be an array but was incorrectly a string in graphql for v1
   class LocationType < Types::BaseObject
+    field :geopoint, String, deprecation_reason: 'Use `geoshape`'
     field :geoshape, String, description: 'GeoShape data for the location, if applicable'
```
JPrevost, line 112: We may want to provide more details on what is in this field. I believe dataeng is consistently (we should confirm) using the BBOX option for the envelope as described here:
https://opensearch.org/docs/latest/field-types/supported-field-types/geo-shape/#envelope

If that turns out to be true, noting something like this for description would probably be helpful.
'GeoShape data for the location when available in WKT format. i.e. `BBOX (minLon, maxLon, maxLat, minLat)`'

matt-bernhardt:
I'd love to write a more helpful description field for `geoshape`, definitely.

JPrevost (approved):
@matt-bernhardt I realize now that we have a bunch of other things that aren't really ready for prod right now until the indexes are updated so holding this one back doesn't make sense. We can merge, and we'll need to carefully plan a release with all of the new geospatial features soon in conjunction with data being released or at least indexes being updated to support the data.

PR #787, opened 2024-02-09
Add support for geodistance search to GraphQL

matt-bernhardt:
### Why are these changes being introduced:

We recently added support for a geodistance search to the OpenSearch model, and now we need to enable that support within GraphQL.

### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/gdt-162

### How does this address that need:

This defines a Geodistance type, with three fields: a distance (as a string), a latitude value (a float), and a longitude value (also a float).

The Geodistance type is then added as an argument of the search field, allowing GraphQL consumers to conduct searches geospatially.

We add three tests to support this functionality:
  - Geospatial search alone
  - Geospatial search combined with a search term
  - Confirming that the geodistance type imposes a requirement for those three fields (by omitting the distance parameter)

### Document any side effects to this change:

* This may be the first time we are using Float as a field type in GraphQL? It seems that we use String elsewhere even when the input is numeric. I'm not sure this is good pracrice, so I wanted to propose that we use a float here.

* There are no tests to confirm that the latitude and longitude fields are required, as it seems unlikely that this requirement would not be met in practice. The distance parameter feels the most likely to be omitted.

#### Developer

#### Code Reviewer

NO

NO


matt-bernhardt:
Closing to re-open - there are issues with the review app, so I'm starting over.

PR #789, opened 2024-02-12
Adds opensearch geo_bounding_box query support

JPrevost:
This has not yet been integrated with GraphQL, so the best way to validate this work is to use the rails console. To test locally, you'll need an opensearch instance running with appropriate data loaded.

```ruby
bin/rails console

q = {geobox: {max_latitude: "42.886",
              min_latitude: "41.239",
              max_longitude: "-73.928",
              min_longitude: "-69.507"}}

Opensearch.new.search(0, q, Timdex::OSClient)
```

You can also combine keyword (or filters) using our existing query syntax:

```ruby
q = {geobox: {max_latitude: "42.886",
              min_latitude: "41.239",
              max_longitude: "-73.928",
              min_longitude: "-69.507"},
       q: "rail stations"}

Opensearch.new.search(0, q, Timdex::OSClient)
```

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/GDT-167

* Adds an optional geo_bounding_box search to our existing opensearch query builder

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/opensearch.rb
+++ b/app/models/opensearch.rb
@@ -111,9 +111,34 @@ class Opensearch
     match_single_field_nested(:subjects, m)
 
     match_geodistance(m) if @params[:geodistance].present?
+    match_geobox(m) if @params[:geobox].present?
     m
   end
 
+  # https://opensearch.org/docs/latest/query-dsl/geo-and-xy/geo-bounding-box/
+  def match_geobox(match_array)
+    match_array << {
+      bool: {
+        must: {
+          match_all: {}
+        },
+        filter: {
+          geo_bounding_box: {
+            'locations.geoshape': {
+              top: @params[:geobox][:max_latitude],
+              bottom: @params[:geobox][:min_latitude],
+              left: @params[:geobox][:min_longitude],
+              right: @params[:geobox][:max_longitude]
+            }
+          }
+        }
+      }
+    }
+  end
+
+  # https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-geo-distance-query.html
+  # Note: at the time of this implementation, opensearch does not have documentation on
+  # this features hence the link to the prefork elasticsearch docs
   def match_geodistance(match_array)
     match_array << {
       bool: {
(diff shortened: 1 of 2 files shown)


test/models/opensearch_test.rb
```diff
@@ -279,6 +279,8 @@ class OpensearchTest < ActiveSupport::TestCase
     os.instance_variable_set(:@params,
                              { geodistance: { latitude: '42.361145', longitude: '-71.057083', distance: '50mi' } })
 
+    refute(os.matches.to_json.include?('{"multi_match":{"query":"rail stations","fields":'))
```
matt-bernhardt, line 282: The intent of this test is to confirm that the query builder isn't hallucinating a keyword search? I think I can understand that motivation, but wonder if there's a smaller string match to look for - like maybe refute on just `multi_match` or something. The way this reads right now, I think hallucinating a search for popcorn would pass the test - as long as it doesn't hallucinate rail stations?

app/models/opensearch.rb
```diff
@@ -111,9 +111,34 @@ def matches
...
+              left: @params[:geobox][:min_longitude],
+              right: @params[:geobox][:max_longitude]
+            }
+          }
+        }
+      }
+    }
+  end
+
+  # https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-geo-distance-query.html
```
matt-bernhardt, line 139: Weird that they haven't created a page for this type of search, but I see what you're talking about.

matt-bernhardt (approved):
I have a fiddly question about the refutation test, but I don't think it is a blocking concern. If you'd rather merge this as-is rather than make a change, I'm fine with that.

:shipi

JPrevost:
@matt-bernhardt I agree with your comment. I added that explicitly to make the pairs of tests more clear as to what was happening in the method being tested between the two, but you are absolutely correct that the `refute` statement is so specific that many other things could be going wrong that it isn't testing. I'll give it peek and see if I can narrow it down in a meaningful way. If not, I may just let it slide but I appreciate pointing it out!

app/models/opensearch.rb
JPrevost: I suspect the Docs licensing didn't allow a fork so I believe OpenSearch had to throw away all the docs and start over whereas with the code they had the fork... so docs required _a lot_ of work to build back up.

test/models/opensearch_test.rb
JPrevost: I narrowed it down and am just squashing/merging without a review as it was so minor. Thanks again.

PR #792, opened 2024-02-14
Add support for geobox searches to GraphQL

matt-bernhardt:
### Why are these changes being introduced:

We have enabled searching via a bounding box within OpenSearch, and need to extend that support to GraphQL in order to make it possible in the UI.

### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/gdt-184

### How does this address that need:

This defines a GeoboxType, with four fields: two latitude and two longitude values, each expressed as floats.

### Document any side effects to this change:

* I've tried to write tests to indicate the edges of what is possible - for example showing that you get results back if you provide both a geodistance and geobox (even if they don't result in overlapping regions - it seems like bbox is used in that case).

* Similarly, I've written a test showing that it is valid to search using either order of the longitude values, because both results are valid - and different - bounding boxes. (If the latitude orders are reversed, the application errors in a way that is not captured via a cassette)

#### Developer

#### Code Reviewer

NO

NO


test/controllers/graphql_controller_v2_test.rb
```diff
@@ -252,6 +252,188 @@ def setup
...
+      assert_equal(200, response.status)
+      json = JSON.parse(response.body)
+
+      assert_nil(json['errors'])
+      assert(json['data']['search']['hits'].positive?)
+    end
+  end
+
+  test 'graphqlv2 geobox search with geodistance search' do
+    # Should this work? It does.
```
JPrevost, line 402: I think a slightly better comment might point out that we don't intend this to be a feature, but we aren't going to prevent people from doing it maybe? I guess we could suggest in the docs that only one geo query at a time is supported.

From my exploration around this I think it ANDs the results together but our data is difficult to reason with at times as we have a lot of things that show up effectively globally that feel like they shouldn't so it's hard to know if our queries are wonky or the data is wonky. For sure I get a smaller set of results when I search for a geobox around Boston and combine with with a geodistance search around the center of India than either of those searches independently which feels correct (even if some of the results are questionable... like "Rivers of Alaska" I'd expect not to be in that search but it is)

JPrevost (approved):
I'd prefer a comment in the tests to be updated to clarify what is going on and how we don't really think people should do that even if we allow it rather than the current comment, but it's not blocking.

test/controllers/graphql_controller_v2_test.rb
matt-bernhardt: I've updated this code comment - will be merging now. Thanks!

PR #795, opened 2024-02-16
Adds Places aggregation

JPrevost:
NOTE: `test/controllers/graphql_controller_v2_test.rb` has significant changes and GitHub's code review windows will not show them by default. Please make sure to review those changes. The cassettes are less interesting and you can probably just confirm one or two indeed scrub the sensitive data but ignore the rest as it was all automated.

* geo ui has reqeusted the ability to filter on additional fields

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/GDT-168
* https://mitlibraries.atlassian.net/browse/GDT-169

* Adds aggregation in OpenSearch for fixed date_ranges
* Adds aggregation in OpenSearch for places by utilizing `Subjects` with a `kind` of `Dublin Core; Spatial`
* Adds GraphQL aggregation capabilities for `date_ranges` and `places`

Not included

* Aggregations for `Access Type` and `Data Type` are not yet included as we are still working out exactly how that data should be stored

The default index we used in tests was changed from the old value to `all-current` which should make it easier to generate new tests and cassettes in the future as the old value (`timdex-prod`) was no longer used anywhere but in our tests.

Changing aggregations required rebuilding most of our cassettes. While I was rebuilding most of them, I opted to rebuild them all. This includes sensitive data scrubbing automatically when using AWS opensearch instances (srubs credentials and URI of the instance). The README has instructions on how to generate cassettes successfully.

I generated cassettes from our current dev1 data, which means a lot of explicitly checked for values changed in the tests. While updating tests for value changes, I also took this opportunity to remove most of the remnants from the v2 naming we had used during the transition from v1. I have not yet renamed the test file, as I was worried that would be more complicated of a review. A follow on change to just rename that file will be forthcoming.

#### Developer

#### Code Reviewer

NO

NO


test/models/opensearch_test.rb
```diff
@@ -67,11 +67,11 @@ class OpensearchTest < ActiveSupport::TestCase
...
   test 'can override index' do
-    # fragile test: assumes opensearch instance with at least one index prefixed with `rdi`
+    # fragile test: assumes opensearch instance with at least one index in the `geo` alias
     VCR.use_cassette('opensearch non-default index') do
-      params = { title: 'data' }
-      results = Opensearch.new.search(0, params, Timdex::OSClient, false, 'rdi*')
-      assert results['hits']['hits'].map { |hit| hit['_index'] }.uniq.map { |index| index.start_with?('rdi') }.any?
+      params = { title: 'bermuda' }
+      results = Opensearch.new.search(0, params, Timdex::OSClient, false, 'geo')
+      assert results['hits']['hits'].map { |hit| hit['_index'] }.uniq.map { |index| index.start_with?('gis') }.any?
```
matt-bernhardt, line 74: I'm not sure that this ultimately matters, but I note that we are querying a `geo` alias and then checking that index values contain not `geo` but `gis`. Part of me wonders whether the test should compare returned results without and with the `geo` argument, looking for a different composition in the search results - but this likely does enough to confirm that when we ask for the geo alias, we get geo records back.

test/controllers/graphql_controller_v2_test.rb
```diff
@@ -1,46 +1,41 @@
 require 'test_helper'
 
 class GraphqlControllerV2Test < ActionDispatch::IntegrationTest
-  def setup
```
matt-bernhardt, line 4: This was a feature removed from `config/features.rb` already, right - so its removal here is more about cleanup than any meaningful change?

test/controllers/graphql_controller_v2_test.rb
```diff
@@ -308,18 +309,19 @@ def setup
...
+    VCR.use_cassette('graphql geobox eastern hemisphere') do
       post '/graphql', params: { query: '{
                                   search(geobox: {
                                     minLongitude: 0,
                                     minLatitude: -90,
                                     maxLongitude: 180,
                                     maxLatitude: 90
-                                  }) {
+                                  },
+                                  source: "MIT GIS Resources") {
```
matt-bernhardt, line 324: I think I understand why we're adding source parameters to this search, because now the tests are running against a bigger index, so a source filter is needed to get down to only GIS records. But if that is the case, the same parameter should also be added to the cassette on line 343 to keep them in sync. Otherwise the test for different results on line 369 can be down to different source values, rather than different geobox definitions.

test/controllers/graphql_controller_v2_test.rb
```diff
@@ -367,8 +369,8 @@ def setup
     refute_equal(eastern_hits, western_hits)
   end
 
-  test 'graphqlv2 geobox search with keyword search' do
-    VCR.use_cassette('graphqlv2 geobox with keyword') do
+  test 'graphql geobox search with keyword search' do
```
matt-bernhardt, line 372: Same question for this test as the other geospatial tests - should we be adding the source parameter here?

test/controllers/graphql_controller_v2_test.rb
```diff
@@ -398,9 +400,9 @@ def setup
     end
   end
 
-  test 'graphqlv2 geobox search with geodistance search' do
+  test 'graphql geobox search with geodistance search' do
```
matt-bernhardt, line 403: Same question for this test as the other geospatial tests - should we be adding the source parameter here?

test/controllers/graphql_controller_v2_test.rb
```diff
@@ -560,9 +561,8 @@ def setup
                  json['errors'].first['message'])
   end
 
-  test 'graphqlv2 valid filters can result in no results' do
-    skip 'opensearch model is not updated to allow this yet'
```
matt-bernhardt, line 564: Nice, I like having fewer skipped tests.

test/controllers/graphql_controller_v2_test.rb
```diff
@@ -632,36 +632,11 @@ def setup
...
-      assert_equal(200, response.status)
-      json = JSON.parse(response.body)
-
-      assert_not json['errors'].present?
-      assert_equal(
-        json['data']['search']['records'].first['locations'][0].keys.sort,
-        ["geopoint", "geoshape", "value", "kind"].sort
-      )
-    end
-  end
```
matt-bernhardt, line 662: Fair enough, yeah - I was iffy on adding this test anyway.

matt-bernhardt (changes requested):
Some of these are questions to make sure I'm not mistaking something, but there's at least one test (the eastern vs western hemisphere box test) which I think needs to be tweaked on order to still be valid. Or maybe the test itself isn't meaningful, as it relies on OpenSearch working correctly and not anything that we're actually doing in the application. Regardless, the change as currently proposes seems to allow for incorrect test passage.

I also have a concern about the year bucketing being applied, because the most recent bucket seems inconsistent with the rest of them.

As you mentioned in the PR, I didn't look at each cassette individually - but did inspect a few of the ones that changed to see how the anonymization was being applied.

I'm happy to chat more about this on Tuesday, though - you're right that there was a lot involved in regenerating all these cassettes, and I'm glad that you took this on.

test/models/opensearch_test.rb
JPrevost: Indexes and aliases for this are different. The alias is indeed `geo`. The indexes start with `gis`. I think this is working as intended even if it isn't entirely clear why.

app/models/aggregations.rb
```diff
@@ -81,4 +85,77 @@ def self.subjects
...
+                from: '2005',
+                to: '2010'
+              },
+              {
+                from: '2010',
+                to: '2015'
+              },
+              {
+                from: '2015',
+                to: 'now/y'
```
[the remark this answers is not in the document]
JPrevost: yeah it sucks honestly like this. Ultimately we haven't added data in so long that these buckets are fairly even in terms of number of results (for MIT GIS data only). I'm definitely open to making a change as this is TIMDEX API wide and not just for GIS. Will adjust.

test/controllers/graphql_controller_v2_test.rb
JPrevost: Good catch. I thought I had done that but clearly didn't. I'll fix.

The reason I am making this change is solely because the test failed with the larger index because the hits for both queries came back as 10,000 so I limited to just MIT data... poorly :)

Will fix.

test/controllers/graphql_controller_v2_test.rb
JPrevost: We shouldn't need to in this case because our assertions don't care if there are more than 10,000 results.

test/controllers/graphql_controller_v2_test.rb
JPrevost: We shouldn't need to in this case because our assertions don't care if there are more than 10,000 results.

test/controllers/graphql_controller_v2_test.rb
JPrevost: I actually don't think I meant to delete this test. At some point I rebased after your code merged and somehow I lost this in the process. I'll reintroduce it.

test/controllers/graphql_controller_v2_test.rb
JPrevost: I've adjusted this test to have both hemispheres use the same source filter. I've also reconfigured it to only use a single cassette for the whole test as we can instruct the cassette to have responses for all of the queries that take place during that test which is generally a bit easier to maintain.

JPrevost:
@matt-bernhardt I think I have addressed and this is ready for another review. Thanks!

app/models/aggregations.rb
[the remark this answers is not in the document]
JPrevost: Noting that I requested a review before adjusting the dates. I'm not sure if I have a great solution yet so I might suggest we land this as-is (as awful as it is) and open a ticket to address improvements. I think we need to start with the UI and design what we want to display and then figure out the best way to achieve that rather than this approach fwiw.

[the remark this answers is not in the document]
JPrevost: Dates filtering is being moved to a separate ticket.

matt-bernhardt (approved):
I think this looks good now - thanks. I reviewed all the tests in the graphql controller file, and I _think_ I get it all.

PR #810, opened 2024-04-04
Add publishers field to record type

jazairi:

We recently [made the decision in transmogrifier](https://github.com/MITLibraries/transmogrifier/pull/132) To map publication information to a new nested field, named `publishers` to avoid a naming conflict with the existing `publicationInformation` field.

#### Relevant ticket(s):

* [GDT-270](https://mitlibraries.atlassian.net/browse/GDT-270)
* [GDT-218](https://mitlibraries.atlassian.net/browse/GDT-218)
* [GDT-229](https://mitlibraries.atlassian.net/browse/GDT-229)
* [GDT-271](https://mitlibraries.atlassian.net/browse/GDT-271)

This adds the new `publishers` field and deprecates `publicationInformation`.

A follow-on ticket is needed to add this field to the UI. (This is ticketed as GDT-271.)

#### Developer

#### Code Reviewer

NO

NO

[GDT-270]: https://mitlibraries.atlassian.net/browse/GDT-270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GDT-218]: https://mitlibraries.atlassian.net/browse/GDT-218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GDT-229]: https://mitlibraries.atlassian.net/browse/GDT-229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GDT-271]: https://mitlibraries.atlassian.net/browse/GDT-271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/graphql/types/record_type.rb
+++ b/app/graphql/types/record_type.rb
@@ -131,6 +131,12 @@ module Types
     end
   end
 
+  class PublishersType < Types::BaseObject
+    field :name, String, description: 'The name of the publisher'
+    field :date, String, description: 'The date of the publication'
+    field :location, String, description: 'Where the publisher is located'
+  end
+
   class RecordType < Types::BaseObject
     field :identifier, ID, null: false, deprecation_reason: 'Use `timdex_record_id`'
     field :timdex_record_id, ID, null: false, description: 'TIMDEX unique identifier for the item'
@@ -158,8 +164,8 @@ module Types
     field :call_numbers, [String], null: true, description: 'Identification number used to classify and locate item'
     field :citation, String, null: true, description: 'Citation for item'
     field :edition, String, null: true, description: 'Edition information for item'
-    field :imprint, [String], null: true, deprecation_reason: 'Use `publicationInformation`'
-    field :publication_information, [String], description: 'Imprint information for item'
+    field :imprint, [String], null: true, deprecation_reason: 'Use `publishers`'
+    field :publication_information, [String], deprecation_reason: 'Use `publishers`'
     field :physical_description, String, null: true, description: 'Physical description of item'
     field :publication_frequency, [String], null: true,
                                             description: 'Publication frequency of item (used for serials)'
@@ -188,6 +194,7 @@ module Types
     field :provider, String,
           null: true,
           description: 'The host institution for a resource. Currently only used for geospatial records'
+    field :publishers, [Types::PublishersType], null: true, description: 'Publishers that are associated with the item'
 
     def in_bibliography
       @object['related_items']&.map { |i| i['uri'] if i['relationship'] == 'IsCitedBy' }&.compact
@@ -203,6 +210,10 @@ module Types
 
     def publication_date
       @object['dates']&.map { |date| date['value'] if date['kind'] == 'Publication date' }&.compact&.first
+    end
+
+    def publication_information
+      @object['publishers']&.map { |publisher| publisher.values.join('; ') }
     end
 
     def file_formats


jazairi:
It's been a minute since I've added a field to the API, so please let me know if I've missed something here (particularly from the testing standpoint)!

JPrevost (commented):
Looks good. Let's confirm with DataEng whether they are mapping _anything_ to the `data` or `location` fields before we merge. If they are not doing that yet, we might be better off removing them for now from the API spec and adding them in when data starts flowing into them.

app/graphql/types/record_type.rb
```diff
@@ -158,8 +164,8 @@ class RecordType < Types::BaseObject
     field :call_numbers, [String], null: true, description: 'Identification number used to classify and locate item'
     field :citation, String, null: true, description: 'Citation for item'
     field :edition, String, null: true, description: 'Edition information for item'
-    field :imprint, [String], null: true, deprecation_reason: 'Use `publicationInformation`'
-    field :publication_information, [String], description: 'Imprint information for item'
+    field :imprint, [String], null: true, deprecation_reason: 'Use `publishers`'
+    field :publication_information, [String], deprecation_reason: 'Use `publishers`'
```
JPrevost, line 168: We should probably do a deprecation method here to populate `publication_information` from a concatenation of the various `publishers` subfields if `publication_information` does not exist. If these words make no sense, let me know and I can better describe what this might look like to allow for the smoothest transition for any consumers of this field.

jazairi: Oh, good call. I've not implemented something like that, but it does make sense. I'm going to check with Graham first to see whether there's data in all subfields before proceeding any further.

JPrevost: Take a look at the methods that follow the RecordType definitions to see how this works. If the method name matches the field, it'll run instead of auto mapping.

So in this case, you'll create a method named `@object['publication_information']` and have it return whatever makes sense. Probably just pass through `publication_information` if it exists (until we fully rip that out of the data which may be soonish?) and construct a string of various concatenated sub fields from `@object['publication']` if not. Or something like that.

JPrevost: Ha you pushed code as I was making that comment. FWIW I think it's probably fine. I don't think `publication_information` is long for this world (I think it goes away after full harvests for each source) so we likely don't need the complexity I suggested.

jazairi: I was just about to ask. :) In that case, I'll re-request review.

app/graphql/types/record_type.rb
```diff
@@ -205,6 +212,10 @@ def publication_date
       @object['dates']&.map { |date| date['value'] if date['kind'] == 'Publication date' }&.compact&.first
     end
 
+    def publication_information
+      @object['publishers']&.map { |publisher| publisher.values.join(', ') }
```
JPrevost, line 216: Would you mind joining on `; `? I worry that a publisher name or location could have commas and be weird-ish.

`Publisher, Inc., December, 1, 1999, Boston, Ma` vs `Publisher, Inc.; December, 1, 1999; Boston, Ma`.

Not a blocker though as both suck and it's a deprecated field lol

JPrevost (approved):
Approved with an alternative separator on the join to consider.

app/graphql/types/record_type.rb
jazairi: Good suggestion. I went ahead and made that change. (Didn't feel like it made sense to request another review, but feel free to confirm if you like!)

PR #811, opened 2024-04-19
Adds support for AND, OR, custom keyword matching

JPrevost:
NOTE: while you can use the GraphQL playground to review these changes, there is also an experimental branch of TIMDEX-UI deployed and configured to use this branch that may also be helpful to more easily see the effects of these proposed changes.

* TIMDEX currently ORs all keyword searches together which leads to a lot of search results, many of which are not relevant (to the extreme of sometimes having no records returned that have all the terms which is sometimes the expectation)

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/GDT-281

* Leverages OpenSearch `minimum_should_match` [feature](https://opensearch.org/docs/latest/query-dsl/minimum-should-match/#valid-values) along with a GraphQL feature to choose a preconfigured value or supply your own

* once we get feedback from stakeholders, it is likely we'll want to adjust the options available, document them more thoroughly in the API docs, and provide them with meaningful names, as well as probably remove some of the options

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -60,7 +60,8 @@ module Types
 
       argument :source, String, required: false, default_value: 'All', deprecation_reason: 'Use `sourceFilter`'
       argument :boolean_type, String, required: false, default_value: 'OR',
-                                      description: 'How to join multiword queries. Defaults to "OR" which means any of the words much match. Options include: "OR", "AND", "minimum_a"'
+                                      description: 'How to join multiword queries. Defaults to "OR" which means any ' \
+                                                   'of the words much match. Options include: "OR", "AND"'
 
       # applied filters
       argument :access_to_files_filter, [String],


app/graphql/types/query_type.rb
```diff
@@ -59,6 +59,8 @@ def record_id(id:, index:)
                                             'you with one for your specific use case'
 
       argument :source, String, required: false, default_value: 'All', deprecation_reason: 'Use `sourceFilter`'
+      argument :boolean_type, String, required: false, default_value: 'OR',
+                                      description: 'How to join multiword queries. Defaults to "OR" which means any of the words much match. Options include: "OR", "AND", "minimum_a"'
```
JPrevost, line 63: Yeah so these docs are already out of date with the code. Will fix before merge.

app/models/opensearch.rb
```diff
@@ -105,6 +105,33 @@ def multisearch
     ]
   end
 
+  # https://opensearch.org/docs/latest/query-dsl/minimum-should-match/#valid-values
+  # checks for preconfigured cases or uses whatever is supplied (i.e. we currently accept OpenSearch syntax for
+  # minimum_should_match)
+  def minimum_should_match
```
JPrevost, line 111: I'm open to holding off on merging until we decide what we want here fwiw. I just needed something in place for stakeholder feedback.

jazairi (commented):
I think this all works as expected, insofar as I trust myself to validate code that involves numbers. I left a comment about the documentation of what I assume became `experiment_a` in the `boolean_type` argument. I'd be fine either with documenting all possible options or just AND and OR (until we decide which experiments will become permanent).

It feels like it would be nice to provide useful information if a bad `boolean_type` were entered. I'm not sure how to achieve that in GraphQL, though, and I don't think that's something we do with other fields.

app/graphql/types/query_type.rb
jazairi: Ha, you beat me to it. Ignore my comment below.

app/models/opensearch.rb
jazairi: That's an interesting idea. As noted below, I have some concerns about introducing/documenting experimental features in the API, since we may remove one or more of them. It's a fairly mild concern, though, since I don't think anyone but us is going to be using these before we decide what we want to do with them.

jazairi (approved):

app/models/opensearch.rb
JPrevost: I opted for just documented AND and OR and merging as that'll leave the timdex-ui experimental PR build working for stakeholders (the API still shuts down PRs after a day). I apologize to our future selves when we forget to cleanup these experiments and can't remember why they are there and aren't documented, etc.

PR #955, opened 2026-04-02
Add support for semantic search

jazairi:

Now that we have the Semantic Query Builder lambda, we need to integrate it into TIMDEX.

#### Relevant ticket(s):

- [USE-411](https://mitlibraries.atlassian.net/browse/USE-411)
- [USE-412](https://mitlibraries.atlassian.net/browse/USE-412)

- Adds SemanticQueryBuilder service to invoke the lambda
- Adds queryMode parameter (keyword/semantic) to GraphQL interface
- Implements builder selection pattern in Opensearch model
- Renames AWS credentials to generic variables (AWS_ACCESS_KEY_ID, etc.)

- Adds aws-sdk-lambda and mocha as dependencies.
- Makes small change to OpenSearch initializer: `ENV.fetch('AWS_SESSION_TOKEN', false)` returns an empty string, which will evaluate to truthy. This has been updated to check for the presence of the env var instead.

#### Developer

- [ ] All new ENV has been added to Heroku Pipeline, Staging and Prod **WARNING: Ensure ENV is updated prior to merge, or builds will break!**

#### Code Reviewer

NO

YES

[USE-411]: https://mitlibraries.atlassian.net/browse/USE-411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[USE-412]: https://mitlibraries.atlassian.net/browse/USE-412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jazairi:
Note that I ran copilot review in a separate draft PR yesterday, to leave us with a cleaner PR for human review. (Copilot had a lot of suggestions.)

app/models/semantic_query_builder.rb
```diff
@@ -0,0 +1,59 @@
...
+  private
+
+  def invoke_semantic_builder(query_text)
+    client_options = {
+      region: ENV.fetch('AWS_REGION', 'us-east-1'),
+      access_key_id: ENV.fetch('AWS_ACCESS_KEY_ID', nil),
+      secret_access_key: ENV.fetch('AWS_SECRET_ACCESS_KEY', nil)
+    }
+
+    client = Aws::Lambda::Client.new(client_options)
```
JPrevost: Did you consider an initializer for the Lambda client like we have for the OpenSearch client? Lambda is weird and different so it may not provide the same persistent connection value we get for OpenSearch, but a shared client would allow for consistency with how we manage those external services. This absolutely might be best here as it only moves a few lines to an initializer and that may end up being more confusing to determine where that is occurring. 🤷🏻

app/models/opensearch.rb
```diff
@@ -54,8 +54,14 @@ def build_query(from)
...
-    @query_strategy ||= LexicalQueryBuilder.new
-    @query_strategy.build(@params, @fulltext)
+    builder = case @query_mode
+              when 'semantic'
+                SemanticQueryBuilder.new
+              else
+                LexicalQueryBuilder.new
+              end
+
+    builder.build(@params, fulltext: @fulltext)
```
JPrevost, line 64: Nice, the pattern we put in place anticipating this is working as cleanly as we hoped.

config/initializers/opensearch.rb
```diff
@@ -15,20 +15,20 @@ def os_client
 def aws_os_client
   OpenSearch::Client.new log: ENV.fetch('OPENSEARCH_LOG', false), url: ENV['OPENSEARCH_URL'] do |config|
     # personal keys use expiring credentials with tokens
-    if ENV.fetch('AWS_OPENSEARCH_SESSION_TOKEN', false)
+    if ENV['AWS_SESSION_TOKEN'].present?
```
JPrevost, line 18: Let's make sure to have all of the affect ENV duplicated in all tiers before this merges please.

test/models/opensearch_test.rb
```diff
@@ -108,4 +108,45 @@ class OpensearchTest < ActiveSupport::TestCase
...
+
+  test 'uses LexicalQueryBuilder by default when no queryMode given' do
+    os = Opensearch.new
+    os.instance_variable_set(:@params, { q: 'test' })
+    os.instance_variable_set(:@fulltext, false)
+    os.instance_variable_set(:@query_mode, nil)
+
+    result = os.query
+
+    # Verify lexical builder returns a bool query
```
JPrevost: Does semantic return a bool as well? I think this an the previous test aren't quite testing what they intend to. I think either asserting LexicalQueryBuilder is called once and/or that SemanticQueryBuilder is not called might be options?

test/models/semantic_query_builder_test.rb
```diff
@@ -0,0 +1,115 @@
...
+        'bool' => {
+          'should' => [
+            { 'rank_feature' => { 'field' => 'embedding_full_record.hello', 'boost' => 6.94 } },
+            { 'rank_feature' => { 'field' => 'embedding_full_record.world', 'boost' => 3.42 } }
+          ]
+        }
+      }
+    }
+
+    mock_lambda_invoke(expected_response)
```
JPrevost: I find this a bit confusing to follow. It isn't clear that what is happening is setting up the next call to the lambda to return what is passed to `mock_lambda_invoke` without a bit of sleuthing.

I wonder if renaming it to `setup_mock_lambda()` would help?

Adding a docstring to the method itself explaining how to use it and what it does might also work.

If I'm correct in what it does this is just conversational and you can choose to make a change or not... if it does something else please help me understand :)

app/models/semantic_query_builder.rb
jazairi: I didn't, but I like that as a way to polish this up a bit!

JPrevost (approved):
Looks good

PR #960, opened 2026-04-23
Add Hybrid Query Builder

jazairi:

Now that we have separate builders for semantic
and lexical queries, we can add an option to
combine the two.

#### Relevant ticket(s):

- [USE-413](https://mitlibraries.atlassian.net/browse/USE-413)
- [USE-414](https://mitlibraries.atlassian.net/browse/USE-414)

This adds a new query builder model for hybrid
queries in OpenSearch, and exposes the new query
type in the GraphQL API.

- We may need to manually weight the scoring due to the different scoring algorithms used by
semantic and lexical search. That felt out of
scope of this change, but may warrant further
exploration.
- The query builder checks only for errors in the semantic builder. We don't have any error
handling in the lexical builder because it
doesn't involve any external calls. This is
probably okay because it's highly unlikely for
lexical search to error, but worth noting.

#### Developer

#### Code Reviewer

NO

NO

[USE-413]: https://mitlibraries.atlassian.net/browse/USE-413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[USE-414]: https://mitlibraries.atlassian.net/browse/USE-414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/hybrid_query_builder.rb
```diff
@@ -0,0 +1,95 @@
...
+    hybrid_bool[:minimum_should_match] = 1 if top_level_filters.present?
+
+    {
+      bool: hybrid_bool
+    }
+  end
+
+  # Logs semantic query builder failures for observability
+  def log_semantic_error(error)
+    return unless defined?(Rails) && Rails.respond_to?(:logger) && Rails.logger
```
JPrevost: This line feels overly cautious? In what situation do we think Rails won't exist and/or the logger won't be configured? We use it all the time in this application without a block like this right?

app/models/hybrid_query_builder.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    Rails.logger.warn(
+      "HybridQueryBuilder semantic query failed: #{error.class}: #{error.message}"
+    )
+  end
+
+  # Only fall back to lexical for Lambda/invocation errors from SemanticQueryBuilder.
+  # All other errors (parsing, validation, unexpected bugs) should be re-raised.
```
JPrevost: I feel like all StandardErrors are raised as a LambdaError from SemanticBuilder. Does this do what the comment says?

app/models/hybrid_query_builder.rb
```diff
@@ -0,0 +1,95 @@
...
+  def semantic_fallback_error?(error)
+    error.is_a?(SemanticQueryBuilder::LambdaError)
+  end
+
+  # Recursively converts all string keys to symbols in hashes and nested structures.
+  # This ensures consistency since semantic builder returns string keys while lexical uses symbols.
```
JPrevost: Is this indicative of an inconsistency we need to fix in one of the Builders rather than normalizing it here? Why are Lexical and Semantic different?

app/models/semantic_query_builder.rb
```diff
@@ -13,16 +16,21 @@ def build(params, fulltext: false)
...
+    function_name = ENV.fetch('TIMDEX_SEMANTIC_BUILDER_FUNCTION_NAME')
+
+    begin
+      response = Timdex::LambdaClient.invoke(
+        function_name: function_name,
+        invocation_type: 'RequestResponse',
+        payload: payload.to_json
+      )
+    rescue StandardError => e
+      # Only Lambda invocation errors are wrapped in LambdaError for graceful fallback
```
JPrevost: I'm curious about this comment and the logic in the Hybrid builder that relies on this. Are we not just raising all standard errors as LambdaErrors here? What does "Only Lambda invocation errors are wrapped in LambdaError for graceful fallback" mean here?

test/models/hybrid_query_builder_test.rb
```diff
@@ -0,0 +1,326 @@
...
+      }
+    }
+
+    lexical_mock = mock
+    lexical_mock.stubs(:build).returns(lexical_result)
+    LexicalQueryBuilder.expects(:new).returns(lexical_mock)
+
+    result = @builder.build(params)
+
+    # When q is blank, should return lexical query (preserves filters)
```
JPrevost, line 24: Does the API actually support this scenario? I think probably?

test/models/hybrid_query_builder_test.rb
```diff
@@ -0,0 +1,326 @@
...
+    lexical_mock = mock
+    lexical_mock.stubs(:build).returns(lexical_result)
+    semantic_mock = mock
+    semantic_mock.stubs(:build).returns(semantic_result)
+    LexicalQueryBuilder.expects(:new).returns(lexical_mock)
+    SemanticQueryBuilder.expects(:new).returns(semantic_mock)
+
+    result = @builder.build(params)
+
+    # Without filters, minimum_should_match should not be set
```
JPrevost, line 203: Can you help me understand why a minimum should match should be different when no filters are in place? I understand why it was in place for the filters, but does it change any results if we kept it for all queries? I'm not asking for change, just help understanding as to what this test

test/models/hybrid_query_builder_test.rb
```diff
@@ -0,0 +1,326 @@
...
+  end
+
+  test 'raises error when semantic builder has non-Lambda error' do
+    params = { q: 'test query' }
+
+    lexical_mock = mock
+    lexical_mock.stubs(:build).returns({ bool: { should: [], must: [], filter: [] } })
+    semantic_mock = mock
+    # Non-Lambda errors (parsing, validation) should be re-raised
+    semantic_mock.stubs(:build).raises(StandardError.new('Invalid semantic query builder response: missing query key'))
```
JPrevost: This is the situation I don't understand if we can actually get to in code even if we can mock it. I may be misinterpreting how it works, but it looks like we just re-raise all standard errors as lambda errors and therefore we can't ever actually get to this outside of mocks.

app/models/hybrid_query_builder.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    # If no query text provided, return lexical query so filters/other constraints are still applied
+    return lexical_query if query_text.blank?
+
+    begin
+      semantic_query = SemanticQueryBuilder.new.build(params, fulltext: fulltext)
+
+      # Both succeeded - combine them with should clause while preserving filters
+      combine_queries(semantic_query, lexical_query)
+    rescue StandardError => e
```
JPrevost: This feels slightly odd to rescue Standard and then filter what we actually want to raise rather than rescuing the actual things we want to handle. Is there a reason you went with StandardError here rather than specific error types?

JPrevost (commented):
I don't really understand some of the implementation decisions and added comments where I could use some more details.

I'd also like confirmation that you manually confirmed expected behavior before generating mocks. Mocks are great if they are used with caution, but with zero tests included that are not mocks I'd like to better understand how much you worked with actual responses in order to create the correct mocks.

I think we are missing GraphQL tests for both Hybrid and Semantic. Maybe if we add those and use VCR instead of mocks I'd feel more comfortable with the mocks here.

app/models/hybrid_query_builder.rb
jazairi: Yeah, I think that's a reasonable conclusion. The issue is that SemanticQueryBuilder receives JSON from the Lambda (which has string keys) and returns it as-is, while LexicalQueryBuilder builds hashes with symbol keys natively. It might be better to have SemanticQueryBuilder return symbol keys. I can make that change.

app/models/hybrid_query_builder.rb
jazairi: Yeah, it does feel overly defensive. We can just rescue `SemanticQueryBuilder::LambdaError` here catching StandardError and filtering.

app/models/hybrid_query_builder.rb
jazairi: Ha! Yeah, if Rails doesn't exist, we have bigger problems. Good catch.

app/models/hybrid_query_builder.rb
jazairi: I think so, yeah. Only the `invoke` call is wrapped in `LambdaError`. `parse_lambda_payload` and `parse_lambda_response` calls happen outside the rescue block, so JSON parsing and response validation errors should propagate as runtime errors.

app/models/semantic_query_builder.rb
jazairi: Yeah, we are. I can see how the comment might suggest we're being more selective. What I meant is that all errors from the `invoke` call get wrapped (allowing graceful fallback), while errors from parsing the response (below the rescue block) don't get wrapped and fail fast. I'll clarify the comment.

test/models/hybrid_query_builder_test.rb
jazairi: It does, yeah. Whether it *should* is another question. :)

test/models/hybrid_query_builder_test.rb
jazairi: It doesn't change anything; it's just redundant to include `minimum_should_match` in non-filter searches. We don't necessarily have to test that behavior if you think it's confusing. (Or, we could change the behavior to always set `minimum_should_match`.)

test/models/hybrid_query_builder_test.rb
jazairi: Yeah, I think you're right. I'll drop this one.

test/models/hybrid_query_builder_test.rb
JPrevost: This makes sense, thanks for helping me understand it.

app/models/hybrid_query_builder.rb
JPrevost: If you'd prefer to treat the SemanticBuilder change as maintenance we can add a ticket or add it to the Confluence page that notes tech debt for this repo. Your call if we do this later or now.

jazairi:
@JPrevost Thanks for the thorough review! Could you let me know if the latest commit addresses your concerns? I added a few VCR tests, but I still stubbed the Lambda responses because I was getting auth errors. I assume it's set up to accept requests only from specific domains, but happy to explore further if that assumption is incorrect.

app/models/hybrid_query_builder.rb
```diff
@@ -12,11 +12,9 @@ def build(params, fulltext: false)
 
       # Both succeeded - combine them with should clause while preserving filters
       combine_queries(semantic_query, lexical_query)
-    rescue StandardError => e
-      # Semantic builder failed (Lambda error, etc.) - use lexical only
+    rescue SemanticQueryBuilder::LambdaError => e
+      # Lambda service failure - gracefully fall back to lexical search
       log_semantic_error(e)
```
JPrevost: I'd bring that logging statement into this block. It currently only does a logging statement so this is mostly just indirection.

JPrevost (commented):
This looks good. Only remaining minor nit is calling a logging method instead of just doing the Rails.logger call directly.

I'm also nervous we're just logging and not throwing to Sentry when we silently fallback like this. That should be considered exceptional so maybe it's worth adding that here too as we'll never notice this in the logs?

No change required, but the lambdas are not restricted from being called from specific URLs so your local .env was likely just not updated to support them. You need the same lambda tier as your open search tier to ensure they work as expected.

JPrevost (approved):

jazairi:
@JPrevost Could you please check the latest commit to wrap the Lambda calls with VCR? It's not a huge change, but I'd like a second set of eyes before we merge.

config/initializers/lambda.rb
```diff
@@ -1,20 +1,17 @@
...
-      session_token: ENV.fetch('AWS_SESSION_TOKEN')
-    )
-  else
-    Aws::Lambda::Client.new(
-      region: ENV.fetch('AWS_REGION', 'us-east-1'),
-      access_key_id: ENV.fetch('AWS_ACCESS_KEY_ID'),
-      secret_access_key: ENV.fetch('AWS_SECRET_ACCESS_KEY')
-    )
-  end
+  options = {
```
JPrevost, line 4: Yeah this is probably better configuration even if you weren't adding the new feature to override the endpoint :)

.env.test
```diff
@@ -2,7 +2,8 @@ EMAIL_FROM=fake@example.com
 EMAIL_URL_HOST=localhost:3000
 JWT_SECRET_KEY=3862fc949629030de4259b88f6e8f7c3702b2fabfc68d00d46fb7f9f70110690b526997ef4d77765ffa010d8aba440286af39947d0c85287174d99be2db14987
 OPENSEARCH_INDEX=all-current
+AWS_ENDPOINT_URL_LAMBDA=http://localhost:9200
```
JPrevost, line 5: I think I'm okay with this not being documented as we don't anticipate it ever being used anywhere but the tests.

JPrevost (approved):

.env.test
jazairi: Oops, I did mean to document it, just to say that it's test env only. Will do so before merging.

PR #964, opened 2026-05-22
Adds use_global_scoring feature

JPrevost:
Adds the ability to use global scoring in OpenSearch queries. This feature allows for more accurate relevance scoring across multiple shards, improving the quality of search results.

This feature comes at a performance cost, as it requires additional coordination between shards. As such, it is disabled by default to ensure only users that require this functionality will enable it.

From the OpenSearch documentation:

```text
dfs_query_then_fetch scores documents using global term and document
frequencies across all shards. It’s usually slower but more accurate.
```

https://mitlibraries.atlassian.net/browse/USE-602
https://mitlibraries.atlassian.net/wiki/spaces/D/pages/5384830981/OpenSearch+Cluster+vs+AOSS+relevancy+differences

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/opensearch.rb
+++ b/app/models/opensearch.rb
@@ -4,15 +4,16 @@ class Opensearch
   MAX_SIZE = 200
 
   def search(from, params, client, highlight: false, index: nil, fulltext: false, query_mode: 'keyword',
-             requested_aggregations: [])
+             requested_aggregations: [], use_global_scoring: false)
     @params = params
     @highlight = highlight
     @fulltext = fulltext?(fulltext)
     @query_mode = query_mode
     @requested_aggregations = requested_aggregations
     index = default_index unless index.present?
-    client.search(index:,
-                  body: build_query(from))
+    search_params = { index:, body: build_query(from) }
+    search_params[:search_type] = 'dfs_query_then_fetch' if use_global_scoring
+    client.search(**search_params)
   end
 
   # Only treat fulltext as true if it is boolean true or the string 'true' (case insensitive)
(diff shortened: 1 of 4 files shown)


app/models/opensearch.rb
```diff
@@ -4,15 +4,16 @@ class Opensearch
   MAX_SIZE = 200
 
   def search(from, params, client, highlight: false, index: nil, fulltext: false, query_mode: 'keyword',
```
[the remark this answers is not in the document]
matt-bernhardt: I'm guessing the alternative here would be to pass these arguments in an argument object or something, but I'm not inclined to worry about this at the moment. There are three required parameters - while optional parameters can increase complexity, I'm more worried about whether the method itself is too complex to work with, rather than what the number of theoretical inputs is.

test/controllers/graphql_controller_test.rb
```diff
@@ -872,7 +872,7 @@ class GraphqlControllerTest < ActionDispatch::IntegrationTest
         initial_hits_count = json_dataset['data']['search']['hits']
         initial_still_images_count = json_dataset['data']['search']['aggregations']['contentType'].find do |x|
                                        x['key'] == 'still image'
-                                     end ['docCount']
+                                     end['docCount']
```
matt-bernhardt, line 875: I'm a little confused about how this doesn't have an impact anywhere, but I see that it doesn't.

matt-bernhardt (approved):
This looks pretty straightforward to me - :shipit:

My comments are non-blocking, and arguably more musings than anything substantive. I see the tests pass, and I see the review app returns different scores (and a slightly different result set towards the bottom of a 20-item response), so this seems useful to have as an option.

test/controllers/graphql_controller_test.rb
JPrevost: It's such a weird syntax anyway that when the linter removed the space I seriously considered refactoring this whole block but then backed away slowly.

PR #978, opened 2026-07-31
adds ADR to expose tuning parameters

JPrevost:


docs/architecture-decisions/0008-expose-tuning-parameters-in-public-graphql-schema.md
```diff
@@ -0,0 +1,203 @@
...
+
+- semantic_must_boost_threshold
+- semantic_drop_boost_threshold
+- semantic_short_query_max_tokens
+
+These arguments are accepted at runtime, but are intentionally hidden from GraphQL introspection by custom visibility behavior in the schema.
+
+This creates a mismatch for downstream clients that validate against introspection-derived schema artifacts, including graphql-client. Those clients fail validation because the schema they consume does not include arguments that our internal callers (timdex-ui in particular) need to send.
+
+Our GraphQL API is public, but exposing advanced tuning arguments is acceptable, even if most external users do not need them. We can state clearly they are likely not useful to most users in the documentation exposed by the Introspection queries.
```
jazairi: I think it would be useful to state clearly why it's acceptable to expose advanced tuning arguments. I assume the worst case scenario is that users who fiddle with the settings might get weird results, which feels low risk.

docs/architecture-decisions/0008-expose-tuning-parameters-in-public-graphql-schema.md
```diff
@@ -0,0 +1,203 @@
...
+The resolver will:
+
+- Accept a structured tuningParameters object.
+- Rely on GraphQL schema validation for shape and field names.
+- Validate value ranges for recognized fields.
+- Pass validated values to semantic_options for semantic and hybrid query execution.
+
+Initial recognized keys:
+
+- mustBoostThreshold (Float, 0.0 to 1.0)
```
jazairi, line 40: Probably not worth mentioning in this doc, as it's an implementation detail, but I don't think we've done this kind of input limitation in TIMDEX before. I'd imagine it would be pretty straightforward.

docs/architecture-decisions/0008-expose-tuning-parameters-in-public-graphql-schema.md
```diff
@@ -0,0 +1,203 @@
...
+
+- Adding new tuning fields still requires schema updates and refreshed client schema artifacts.
+- Requires range validation tests for semantic thresholds and token limits.
+- If we want tuning knobs in the future that we aren't comfortable making public, we would need to consider another input object closer to Option C
+
+## Implementation Plan
+
+Tips:
+
+- Add test coverage for accepted and invalid tuning values.
```
jazairi, line 194: Ignore my comment above about input validation. I think this addresses it, since test coverage would imply existing validation code.

jazairi (approved):
I support option B. I think strict typing is critical here. As I understand it, option C would see any JSON object and validate it based on data type alone. Even if it's probably fine in practice, delegating validation elsewhere in the codebase feels risky to me, and not worth the perceived benefits.

docs/architecture-decisions/0008-expose-tuning-parameters-in-public-graphql-schema.md
JPrevost: Great point. Yes, the risk for all of our current tuning parameters is excluding results that would have been useful or returning way more irrelevant results after the main good results. I'll update to clarify explicitly why this is low risk as you suggested.

matt-bernhardt (approved):
I support option B, for all the stated reasons in the document and the comments here. Having these included in the public schema should prove vastly more workable when it comes to actually using these parameters during measurement and tuning operations.

(I went down a bit of a rabbit hole trying to imagine how these different options might work from the Quepid interface before realizing that the UI application will be the thing mediating between the querystring which Quepid exposes and the GraphQL syntax that is envisioned here - as evidenced by how we already handle this in a GeoData context for the GeoBox and GeoDistance constructs.)

docs/architecture-decisions/0008-expose-tuning-parameters-in-public-graphql-schema.md
JPrevost: Updated with wording clarifying why it is okay to expose our tuning parameters.

